PR #9, opened 2026-03-11
Add benchmark runner and convergence proof

datablogin:
## Summary

- **BenchmarkRunner** (`causal_optimizer/benchmarks/runner.py`): Runs benchmark SCMs with three strategies — `causal` (ExperimentEngine + causal graph), `random` (uniform sampling), and `surrogate_only` (ExperimentEngine without graph). Supports `run()` for single runs and `compare()` for multi-seed comparison. Returns `BenchmarkResult` with convergence curves, final best, and experiments-to-threshold.
- **HighDimensionalSparseBenchmark** (`causal_optimizer/benchmarks/high_dimensional.py`): 20-variable benchmark where only 3 (X1→X2→X3→objective chain) affect the outcome. Tests that causal-aware optimization can ignore irrelevant dimensions.
- **Convergence proof test** (`tests/integration/test_causal_beats_naive.py`): Directional test showing causal strategy outperforms random on ToyGraphBenchmark (budget=30, n_seeds=3).

## Test plan

- [ ] 29 new integration tests in `tests/integration/test_benchmarks.py` covering:
  - HighDimensionalSparseBenchmark (search space, graph, POMIS, run, irrelevant var isolation, determinism)
  - BenchmarkResult dataclass fields
  - Runner smoke tests: all 4 benchmarks × 3 strategies (budget=5)
  - Convergence curve monotonicity for all strategies
  - `compare()` returns correct result counts and seeds
  - Edge cases: invalid strategy, budget=1, threshold calculation
- [ ] 1 slow test in `tests/integration/test_causal_beats_naive.py`: causal vs random directional comparison
- [ ] All 153 tests pass (152 non-slow + 1 slow)
- [ ] `ruff check` clean, `ruff format` clean, `mypy` strict clean

<h3>Greptile Summary</h3>

This PR introduces a `BenchmarkRunner` that compares three optimization strategies (`causal`, `random`, `surrogate_only`) across any `BenchmarkSCM` implementation, adds a new `HighDimensionalSparseBenchmark` with 20 variables and only 3 causal ancestors, and provides a directional convergence proof test backed by 29 integration tests.

The implementation is well-structured and the design trade-offs (e.g. acknowledged non-reproducibility of the engine's internal RNG, intervention-dependent RNG stream position) are clearly documented. One correctness issue was found:

- **`final_best` type leak** (`runner.py` lines 196, 243): Both `_run_engine` and `_run_random` initialize `best_so_far = float("inf")`, but after the first `min(float_inf, numpy_float64)` call the value silently becomes `numpy.float64`. This violates the `BenchmarkResult.final_best: float` annotation and causes `isinstance(result.final_best, float)` in the smoke test (line 131) to fail under numpy ≥ 2.0, where `np.float64` no longer subclasses `float`. The fix is to cast with `float(best_so_far)` when constructing `BenchmarkResult`.
- **Test assertion consistency** (`test_benchmarks.py` line 131): The bare `isinstance(result.final_best, float)` check should use `(float, np.floating)` to match the pattern already applied at line 59.
- **Missing pre-validation of `budget` in `compare()`**: All other parameters are validated up-front; `budget` is only caught lazily on the first `run()` call.

<h3>Confidence Score: 3/5</h3>

- Safe to merge after fixing the `numpy.float64` / `float` type leak in `BenchmarkResult.final_best`.
- Core logic is sound and well-tested, but the `final_best` type inconsistency is a concrete runtime failure: the smoke test assertion `isinstance(result.final_best, float)` will break on any environment running numpy 2.0+, which is within the unpinned `numpy>=1.24.0` range declared in `pyproject.toml`. All other issues are minor (test assertion style, validation ordering).
- `causal_optimizer/benchmarks/runner.py` (both `_run_engine` and `_run_random` return `final_best` as `numpy.float64`); `tests/integration/test_benchmarks.py` line 131 (bare `isinstance(..., float)` check).

| Filename | Overview |
|----------|----------|
| causal_optimizer/benchmarks/runner.py | New benchmark runner with three strategies (causal, random, surrogate_only); contains a type consistency bug where `final_best` can be `numpy.float64` instead of Python `float`, breaking numpy 2.0+ compatibility and violating the dataclass annotation. |
| causal_optimizer/benchmarks/high_dimensional.py | New 20-variable sparse benchmark with a 3-node causal chain; structural equations are correct, docstring now accurately documents that RNG stream position is intervention-dependent. |
| tests/integration/test_benchmarks.py | 29 integration tests with good coverage; line 131 retains a bare `isinstance(result.final_best, float)` check that will fail under numpy 2.0+, while line 59 was already fixed to use `(float, np.floating)`. |
| tests/integration/test_causal_beats_naive.py | Directional convergence proof test with generous tolerance and well-documented stochasticity caveats; both methods correctly inherit the `@pytest.mark.slow` marker from the class. |
| causal_optimizer/benchmarks/__init__.py | Adds `HighDimensionalSparseBenchmark`, `BenchmarkResult`, `BenchmarkRunner` to the public API and adds `noise_scale: float` to the `BenchmarkSCM` protocol; changes are clean and consistent. |
| pyproject.toml | Adds the `slow` pytest marker; straightforward and correct. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Caller
    participant R as BenchmarkRunner
    participant RE as _run_engine
    participant RR as _run_random
    participant E as ExperimentEngine
    participant B as BenchmarkSCM

    C->>R: compare(strategies, budget, n_seeds)
    R->>R: validate n_seeds, strategies
    loop for each (strategy, seed)
        R->>R: run(strategy, budget, seed)
        alt strategy == "random"
            R->>RR: _run_random(budget, seed, ...)
            RR->>RR: SeedSequence.spawn() → noise_rng, sample_rng
            RR->>B: new benchmark(noise_scale, noise_rng)
            loop budget times
                RR->>RR: _sample_random_params(space, sample_rng)
                RR->>B: bench.run(params)
                B-->>RR: {"objective": numpy.float64}
                RR->>RR: best_so_far = min(best_so_far, obj)
            end
            RR-->>R: BenchmarkResult(final_best=best_so_far)
        else strategy == "causal" or "surrogate_only"
            R->>RE: _run_engine(strategy, budget, seed, ...)
            RE->>B: new benchmark(noise_scale, rng=default_rng(seed))
            RE->>E: ExperimentEngine(space, bench, graph?)
            loop budget times
                RE->>E: engine.step()
                E->>B: bench.run(suggested_params)
                B-->>E: metrics
                E-->>RE: ExperimentResult
                RE->>RE: best_so_far = min(best_so_far, obj)
            end
            RE-->>R: BenchmarkResult(final_best=best_so_far)
        end
    end
    R-->>C: list[BenchmarkResult]
```

<details open><summary><h3>Comments Outside Diff (6)</h3></summary>

1. `causal_optimizer/benchmarks/runner.py`, line 324-328 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/causal_optimizer/benchmarks/runner.py#L324-L328)) 

   **`ExperimentEngine` created without a seed — engine-based strategies are non-deterministic**

   `ExperimentEngine.__init__` accepts no `seed` or `rng` parameter (the constructor signature is `search_space, runner, objective_name, minimize, causal_graph, max_skips, descriptor_names`). This means the engine's internal RNG — which governs experiment suggestions, screening, bootstrap, and off-policy prediction — is freshly initialised from system entropy on every call to `_run_engine`.

   Concretely, two back-to-back calls to `runner.run("causal", budget=5, seed=0)` will almost always produce different `convergence_curve` values, because the benchmark noise RNG is deterministic (seeded from `seed`) but the *experiments that are run* vary randomly.

   Practical impact:
   1. `compare()` will give different numerical results on every test run for the `"causal"` and `"surrogate_only"` strategies.
   2. The directional test `test_causal_vs_random_toy_graph` / `test_causal_vs_random_high_dimensional` relies on the causal advantage being large enough to overcome this variance across only 3 seeds.

   If `ExperimentEngine` gains a seed parameter in future, it should be threaded through here:

   ```python
   engine = ExperimentEngine(
       search_space=space,
       runner=bench,
       causal_graph=graph,
       # seed=seed,  # add when ExperimentEngine supports it
   )
   ```

   At minimum, a comment near the `seed` parameter in `run()` should call this out explicitly, since the docstring currently only says "the engine's internal RNG (suggestions, bootstrap) is not controlled" without explaining that this means the `seed` argument has *no effect* on reproducibility for engine-based strategies.

2. `tests/integration/test_causal_beats_naive.py`, line 789-795 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/tests/integration/test_causal_beats_naive.py#L789-L795)) 

   **Tolerance of `+1.0` is very weak relative to the objective scale**

   For `ToyGraphBenchmark`, the negated objective is `-(cos(z) - exp(-z/20))`. The practical range of `final_best` is roughly `[-1.2, +1.2]`, so a tolerance of `1.0` spans nearly the full dynamic range of the benchmark. Concretely, this assertion passes even if the causal strategy achieves `avg_causal = 0.0` while random achieves `avg_random = -1.0` — i.e., the causal strategy performs at the *worst* possible outcome while random finds the optimum.

   Combined with the non-deterministic engine (see comment on `_run_engine`), the test could pass even when the causal strategy is completely broken, as long as random also happens to perform poorly.

   Consider tightening to 10–20% of the objective scale, or expressing the assertion as a fraction of `avg_random`:

   ```python
   # Causal should converge at least as well as random (within noise).
   assert avg_causal <= avg_random + 0.2, ...
   ```

   Alternatively, increase `n_seeds` (e.g., to 5 or 10) to reduce variance before applying a tighter tolerance. The test docstring already notes this option.

3. `causal_optimizer/benchmarks/runner.py`, line 419-438 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/causal_optimizer/benchmarks/runner.py#L419-L438)) 

   **`_sample_random_params` always samples ALL variables including irrelevant ones**

   For `HighDimensionalSparseBenchmark`, this function samples all 20 variables (x1–x20) and passes them all to `bench.run()`. The benchmark's `run()` method silently ignores x4–x20 because they have no causal effect on the objective.

   This creates an asymmetry between strategies:
   - `"random"`: intervenes on x1, x2, and x3 simultaneously (overriding all structural equations)
   - `"causal"` (POMIS = `{x3}`): intervenes only on x3, allowing x1 and x2 to follow their structural equations

   So "random" doesn't just waste budget on irrelevant variables — it also changes the semantics of the causal chain by overriding x1 and x2 structurally. The causal strategy explores x3 optimally *conditioned on the structural propagation* from x1→x2→x3, while random explores x3 directly (but also x1 and x2 uncorrelated).

   This is arguably by design (random is supposed to be naive), but it's worth a comment to clarify:

   ```python
   # NOTE: for benchmarks with a causal chain (e.g. HighDimensionalSparseBenchmark),
   # this samples all variables including upstream ones, effectively intervening on the
   # full chain rather than just the target node. This is intentional — random does not
   # exploit structural equations.
   params = _sample_random_params(space, sample_rng)
   ```

4. `causal_optimizer/benchmarks/runner.py`, line 192-201 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/causal_optimizer/benchmarks/runner.py#L192-L201)) 

   **`final_best` silently becomes `numpy.float64` instead of `float`**

   `obj` is sourced from `result.metrics.get("objective", float("inf"))`. When the benchmark's `run()` returns `numpy.float64` scalars (e.g. from `np.cos(...)` or `rng.normal(...)`), `min(float("inf"), numpy.float64(...))` returns `numpy.float64`. `best_so_far` therefore becomes `numpy.float64` after the first step, violating the `final_best: float` annotation on `BenchmarkResult`.

   This manifests concretely in the smoke test at `tests/integration/test_benchmarks.py:131`:
   ```python
   assert isinstance(result.final_best, float)
   ```
   In numpy ≥ 2.0, `np.float64` no longer subclasses Python `float`, so that assertion fails. The same issue exists in `_run_random` at line 243.

   Fix: cast explicitly when constructing `BenchmarkResult`:
   ```python
           return BenchmarkResult(
               convergence_curve=convergence_curve,
               final_best=float(best_so_far),
               experiments_to_threshold=experiments_to_threshold,
               strategy=strategy,
               benchmark_name=benchmark_name,
               seed=seed,
           )
   ```
   Apply the same `float(best_so_far)` cast in `_run_random`.

5. `tests/integration/test_benchmarks.py`, line 131 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/tests/integration/test_benchmarks.py#L131)) 

   **`isinstance(result.final_best, float)` fragile under numpy 2.0+**

   `_run_engine` and `_run_random` both produce a `best_so_far` that can be `numpy.float64` (see comment on `runner.py`). With numpy ≥ 2.0, `np.float64` no longer inherits from Python `float`, so this assertion would fail even though the value is semantically correct.

   The same file already handles this correctly at line 59 using `(float, np.floating)`. Either apply the same pattern here, or fix the root cause in the runner by casting with `float(best_so_far)` before storing in `BenchmarkResult`.

6. `causal_optimizer/benchmarks/runner.py`, line 147-155 ([link](https://github.com/datablogin/causal-optimizer/blob/0333ab9cc7f2cf1a3a5a38003ef0dc404ff7b2ee/causal_optimizer/benchmarks/runner.py#L147-L155)) 

   **`compare()` does not validate `budget` before iterating**

   `compare()` validates `n_seeds` and all strategy names up-front, but does not check `budget`. If a caller passes `budget=0`, the guard in `run()` will fire, but only after the strategy/seed loop has already started (i.e., the error is raised on the first iteration rather than at the call site). This is a minor inconsistency: all other parameters are pre-validated, but `budget` is not.

   Consider mirroring the `n_seeds` guard:
   ```python
   if budget <= 0:
       msg = f"budget must be a positive integer, got {budget!r}."
       raise ValueError(msg)
   ```

</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%203%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%203%0Acausal_optimizer%2Fbenchmarks%2Frunner.py%3A192-201%0A**%60final_best%60%20silently%20becomes%20%60numpy.float64%60%20instead%20of%20%60float%60**%0A%0A%60obj%60%20is%20sourced%20from%20%60result.metrics.get%28%22objective%22%2C%20float%28%22inf%22%29%29%60.%20When%20the%20benchmark's%20%60run%28%29%60%20returns%20%60numpy.float64%60%20scalars%20%28e.g.%20from%20%60np.cos%28...%29%60%20or%20%60rng.normal%28...%29%60%29%2C%20%60min%28float%28%22inf%22%29%2C%20numpy.float64%28...%29%29%60%20returns%20%60numpy.float64%60.%20%60best_so_far%60%20therefore%20becomes%20%60numpy.float64%60%20after%20the%20first%20step%2C%20violating%20the%20%60final_best%3A%20float%60%20annotation%20on%20%60BenchmarkResult%60.%0A%0AThis%20manifests%20concretely%20in%20the%20smoke%20test%20at%20%60tests%2Fintegration%2Ftest_benchmarks.py%3A131%60%3A%0A%60%60%60python%0Aassert%20isinstance%28result.final_best%2C%20float%29%0A%60%60%60%0AIn%20numpy%20%E2%89%A5%202.0%2C%20%60np.float64%60%20no%20longer%20subclasses%20Python%20%60float%60%2C%20so%20that%20assertion%20fails.%20The%20same%20issue%20exists%20in%20%60_run_random%60%20at%20line%20243.%0A%0AFix%3A%20cast%20explicitly%20when%20constructing%20%60BenchmarkResult%60%3A%0A%60%60%60python%0A%20%20%20%20%20%20%20%20return%20BenchmarkResult%28%0A%20%20%20%20%20%20%20%20%20%20%20%20convergence_curve%3Dconvergence_curve%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20final_best%3Dfloat%28best_so_far%29%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20experiments_to_threshold%3Dexperiments_to_threshold%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20strategy%3Dstrategy%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20benchmark_name%3Dbenchmark_name%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20seed%3Dseed%2C%0A%20%20%20%20%20%20%20%20%29%0A%60%60%60%0AApply%20the%20same%20%60float%28best_so_far%29%60%20cast%20in%20%60_run_random%60.%0A%0A%23%23%23%20Issue%202%20of%203%0Atests%2Fintegration%2Ftest_benchmarks.py%3A131%0A**%60isinstance%28result.final_best%2C%20float%29%60%20fragile%20under%20numpy%202.0%2B**%0A%0A%60_run_engine%60%20and%20%60_run_random%60%20both%20produce%20a%20%60best_so_far%60%20that%20can%20be%20%60numpy.float64%60%20%28see%20comment%20on%20%60runner.py%60%29.%20With%20numpy%20%E2%89%A5%202.0%2C%20%60np.float64%60%20no%20longer%20inherits%20from%20Python%20%60float%60%2C%20so%20this%20assertion%20would%20fail%20even%20though%20the%20value%20is%20semantically%20correct.%0A%0AThe%20same%20file%20already%20handles%20this%20correctly%20at%20line%2059%20using%20%60%28float%2C%20np.floating%29%60.%20Either%20apply%20the%20same%20pattern%20here%2C%20or%20fix%20the%20root%20cause%20in%20the%20runner%20by%20casting%20with%20%60float%28best_so_far%29%60%20before%20storing%20in%20%60BenchmarkResult%60.%0A%0A%60%60%60suggestion%0A%20%20%20%20%20%20%20%20assert%20isinstance%28result.final_best%2C%20%28float%2C%20np.floating%29%29%0A%60%60%60%0A%0A%23%23%23%20Issue%203%20of%203%0Acausal_optimizer%2Fbenchmarks%2Frunner.py%3A147-155%0A**%60compare%28%29%60%20does%20not%20validate%20%60budget%60%20before%20iterating**%0A%0A%60compare%28%29%60%20validates%20%60n_seeds%60%20and%20all%20strategy%20names%20up-front%2C%20but%20does%20not%20check%20%60budget%60.%20If%20a%20caller%20passes%20%60budget%3D0%60%2C%20the%20guard%20in%20%60run%28%29%60%20will%20fire%2C%20but%20only%20after%20the%20strategy%2Fseed%20loop%20has%20already%20started%20%28i.e.%2C%20the%20error%20is%20raised%20on%20the%20first%20iteration%20rather%20than%20at%20the%20call%20site%29.%20This%20is%20a%20minor%20inconsistency%3A%20all%20other%20parameters%20are%20pre-validated%2C%20but%20%60budget%60%20is%20not.%0A%0AConsider%20mirroring%20the%20%60n_seeds%60%20guard%3A%0A%60%60%60python%0Aif%20budget%20%3C%3D%200%3A%0A%20%20%20%20msg%20%3D%20f%22budget%20must%20be%20a%20positive%20integer%2C%20got%20%7Bbudget!r%7D.%22%0A%20%20%20%20raise%20ValueError%28msg%29%0A%60%60%60%0A%0A&repo=datablogin%2Fcausal-optimizer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/badge/Fix_All_in_Claude-black?logo=claude&logoColor=%23D97706">Fix All in Claude Code</picture></a>

<sub>Last reviewed commit: 0333ab9</sub>

### DIFF
--- a/causal_optimizer/benchmarks/runner.py
+++ b/causal_optimizer/benchmarks/runner.py
@@ -137,8 +137,13 @@ class BenchmarkRunner:
             List of ``BenchmarkResult``, one per (strategy, seed) pair.
 
         Raises:
-            ValueError: If any strategy name is not supported.
+            ValueError: If any strategy name is not supported, or if ``n_seeds``
+                is not a positive integer.
         """
+        if n_seeds <= 0:
+            msg = f"n_seeds must be a positive integer, got {n_seeds!r}."
+            raise ValueError(msg)
+
         for strategy in strategies:
             self._validate_strategy(strategy)
 
--- a/tests/integration/test_benchmarks.py
+++ b/tests/integration/test_benchmarks.py
@@ -56,7 +56,7 @@ class TestHighDimensionalBenchmark:
         bench = HighDimensionalSparseBenchmark(rng=np.random.default_rng(42))
         result = bench.run({"x1": 1.0, "x2": 0.5, "x3": 0.1})
         assert "objective" in result
-        assert isinstance(result["objective"], float)
+        assert isinstance(result["objective"], (float, np.floating))
 
     def test_irrelevant_variables_ignored(self) -> None:
         """Changing x4-x20 should not affect the objective (noise aside)."""
@@ -199,6 +199,18 @@ class TestRunnerEdgeCases:
         runner = BenchmarkRunner(bench)
         with pytest.raises(ValueError, match="Unknown strategy"):
             runner.run(strategy="invalid", budget=5)
+
+    def test_zero_n_seeds_raises(self) -> None:
+        bench = ToyGraphBenchmark(rng=np.random.default_rng(42))
+        runner = BenchmarkRunner(bench)
+        with pytest.raises(ValueError, match="n_seeds must be a positive integer"):
+            runner.compare(strategies=["random"], budget=5, n_seeds=0)
+
+    def test_negative_n_seeds_raises(self) -> None:
+        bench = ToyGraphBenchmark(rng=np.random.default_rng(42))
+        runner = BenchmarkRunner(bench)
+        with pytest.raises(ValueError, match="n_seeds must be a positive integer"):
+            runner.compare(strategies=["random"], budget=5, n_seeds=-1)
 
     def test_budget_one(self) -> None:
         bench = ToyGraphBenchmark(rng=np.random.default_rng(42))


datablogin:
# Claude Code Review

## Review Feedback

Good, I have a thorough understanding of all the code. Here's my review:

---

## PR Review: Add benchmark runner and convergence proof

### Overall Assessment

Well-structured PR that adds a benchmark runner framework and a high-dimensional sparse benchmark. The code is clean, well-documented, and follows existing patterns. The test coverage is thorough. A few issues worth addressing:

---

### Issues

#### 1. Bug: `_run_engine` doesn't pass `objective_name` or `minimize` to `ExperimentEngine`

`runner.py:151-155` — The engine is constructed without `objective_name` or `minimize`, relying on the defaults (`"objective"`, `True`). This happens to be correct for all current benchmarks, but the runner docstring says it "hardcodes" this, and the TODO comments acknowledge it. Fine for now, but note that if a benchmark ever uses a different metric key or maximization, this silently does the wrong thing. The TODO comments are good — just flagging that this is a latent coupling.

#### 2. Bug: `rng` created but unused in `run()` for engine strategies

`runner.py:101` — `rng = np.random.default_rng(seed)` is created in `run()` but only used for the `"random"` strategy path. For `"causal"` and `"surrogate_only"`, it's a wasted allocation. Minor, but slightly misleading — a reader might think the engine strategies use this rng.

#### 3. Potential issue: `_run_random` uses two different RNG streams with correlated seeds

`runner.py:183-191` — The benchmark gets `rng=seed + 1_000_000` for structural noise while sampling uses `rng=seed` (passed in from `run()`). The docstring explains this is intentional for decorrelation, which is fine. But the seed offset of 1M is a magic number — if someone passes `seed=1_000_000`, the benchmark's noise RNG gets `seed=2_000_000`, which is fine, but `seed=-1_000_000` would give `seed=0`. No actual bug, just noting the convention could be fragile for negative seeds. Consider documenting the valid seed range or using `SeedSequence.spawn()` for proper stream splitting:

```python
ss = np.random.SeedSequence(seed)
sample_rng, noise_rng = [np.random.default_rng(s) for s in ss.spawn(2)]
```

#### 4. `BenchmarkRunner.__init__` types `self.benchmark` as `Any`

`runner.py:71` — The constructor accepts `BenchmarkSCM` but immediately casts to `Any`. This defeats type checking for all downstream usage (`.noise_scale`, `.run()`, etc.). The reason is clear — `BenchmarkSCM` protocol doesn't declare `noise_scale` — but a better approach would be to keep the protocol type and access `noise_scale` via `getattr` or add `noise_scale` to the protocol:

```python
class BenchmarkSCM(Protocol):
    noise_scale: float  # add this
    ...
```

This would give you proper type safety everywhere instead of escaping to `Any`.

#### 5. `_run_engine` relies on `type(self.benchmark)` constructor convention

`runner.py:143-147` — The runner creates fresh benchmark instances via `benchmark_type(noise_scale=..., rng=...)`. This assumes all benchmarks have this exact constructor signature. The Protocol can't enforce this (as the docstring notes). This is a known limitation, but worth noting: a benchmark that adds required constructor args (e.g., `n_variables`) would silently break.

Consider documenting this constructor contract more prominently, or adding a factory method to the protocol like `create(noise_scale, rng) -> Self`.

#### 6. `high_dimensional.py` — SCM noise is consumed even when variables are intervened on

`high_dimensional.py:75-78` — When `x2` is provided in `parameters`, the code still computes `x2_structural` (consuming an RNG draw). Same for `x3`. This means the RNG state diverges depending on which parameters are intervened on, which is correct for partial intervention semantics (the noise term for unmeasured variables should still advance the RNG). Just noting this is a deliberate choice that matches the existing ToyGraph pattern.

#### 7. Test flakiness risk in `test_causal_beats_naive.py`

`test_causal_beats_naive.py:50-55` — The tolerance of `avg_causal <= avg_random + 1.0` is quite generous. With `budget=30` and `n_seeds=3`, this is really testing "causal doesn't catastrophically fail" rather than "causal outperforms random." The docstring is honest about this, which is good.

However, with only 3 seeds and 30 experiments each, the variance can be substantial. If this test becomes flaky, the fix should be increasing `n_seeds` (not `budget`), since more seeds reduce the variance of the mean more effectively. The docstring suggests increasing budget, but that only helps if per-run variance is the issue.

#### 8. Missing validation: `compare()` doesn't validate strategy names

`runner.py:108-131` — `compare()` delegates to `run()` which validates, so this is fine functionally. But if invalid strategy names are passed, you get one error at a time rather than failing fast. Consider validating all strategies upfront:

```python
invalid = set(strategies) - self._STRATEGIES
if invalid:
    raise ValueError(...)
```

---

### Positive observations

- **Excellent documentation**: Docstrings are thorough, TODOs are well-placed and reference CLAUDE.md known issues, and the test file's module-level docstring in `test_causal_beats_naive.py` is a model of how to document flaky-by-nature statistical tests.

- **Clean protocol usage**: `BenchmarkSCM` protocol is well-designed. The `@staticmethod` methods for `search_space()`, `causal_graph()`, `known_pomis()` are the right call since they don't depend on instance state.

- **`_sample_random_params` handles all variable types**: CONTINUOUS, INTEGER, BOOLEAN, CATEGORICAL all covered with proper bounds handling and error messages.

- **Good test structure**: The parametrized smoke tests (`TestRunnerSmoke`) cover the full cross-product of benchmarks x strategies. The threshold computation tests (`test_threshold_step_exact`, `test_threshold_step_zero_optimum`) test edge cases directly.

- **`known_pomis` validated against algorithm**: `test_known_pomis_matches_algorithm` is an excellent test — it verifies the hardcoded POMIS matches `compute_pomis()` output, catching drift between the benchmark and the algorithm.

- **`@pytest.mark.slow` and marker registration**: The slow marker is properly registered in `pyproject.toml` with a clear deselect hint.

---

### Minor/style

- `high_dimensional.py:45` — `HighDimensionalSparseBenchmark.N_VARIABLES` in the classmethod body is a bit verbose. Could use `cls` if converted to `@classmethod`, but since it's `@staticmethod` this is the correct way. No change needed.

- `runner.py:228` — `if known_optimum != 0.0` uses float equality, which is technically fragile but fine here since the caller controls the value. `math.isclose` would be pedantic but not necessary.

---

### Summary

This is a solid PR. The main actionable items are:

1. **Add `noise_scale: float` to the `BenchmarkSCM` protocol** to avoid the `Any` escape hatch.
2. **Consider `SeedSequence.spawn()`** instead of the magic seed offset for RNG decorrelation.
3. **Validate strategies upfront** in `compare()`.

Everything else is either acknowledged via TODOs or is a reasonable design choice given the current state of the codebase.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the codebase. Here's my review:

---

## PR Review: Add benchmark runner and convergence proof

### Overall Assessment

This is a well-structured PR that adds meaningful infrastructure: a high-dimensional benchmark for stress-testing causal guidance, a general-purpose benchmark runner for strategy comparison, and solid test coverage. The code is clean, well-documented, and follows existing patterns. I have a handful of findings ranging from a potential bug to minor suggestions.

---

### Issues

**1. RNG stream not split in `_run_random` — `rng` parameter is shadowed and unused**
`runner.py:107-109`

In `run()`, a fresh `rng` is created from `seed` and passed to `_run_random`. But inside `_run_random`, a *new* `SeedSequence(seed)` is created and spawned for the noise RNG, while the `rng` parameter passed in is never actually used for sampling — it's the one used by `_sample_random_params`. So the flow works, but there's an inconsistency: the `rng` created at line 108 and the `noise_rng` spawned at line 204 are *not* properly decorrelated because they're derived from the same seed differently (`default_rng(seed)` vs `SeedSequence(seed).spawn()`). The comment says "one stream for benchmark structural noise, one for parameter sampling" but the parameter sampling stream (`rng`) was created independently at the call site, not via `spawn()`.

**Fix:** Create both RNG streams from the same `SeedSequence.spawn()` call inside `_run_random`:
```python
ss = np.random.SeedSequence(seed)
child_seeds = ss.spawn(2)
noise_rng = np.random.default_rng(child_seeds[0])
sample_rng = np.random.default_rng(child_seeds[1])
```
Then use `sample_rng` instead of the `rng` parameter, or remove the `rng` parameter entirely.

**2. `_run_engine` doesn't handle CRASH results**
`runner.py:173-179`

When `engine.step()` returns a CRASH result, the engine sets `objective` to `float("inf")` (per `engine/loop.py` crash handling). The runner then picks this up via `metrics.get("objective", float("inf"))` — so crashes don't pollute `best_so_far`. This works correctly for the convergence curve, but it means a CRASH still counts toward the `budget`. This is arguably correct (a crashed experiment still costs budget), but worth documenting explicitly since it's a design choice.

**3. `_run_engine` doesn't pass `objective_name` or `minimize` to `ExperimentEngine`**
`runner.py:164-168`

The `ExperimentEngine` constructor accepts `objective_name` and `minimize` parameters, but the runner always uses the defaults (`"objective"` and `True`). This is consistent with the TODO comments and the known `best_result` bug, but it means the runner is tightly coupled to the `"objective"` key convention. The TODOs document this well — just flagging for completeness.

**4. Protocol adds `noise_scale` attribute requirement — verify all implementations**
`__init__.py:27`

The PR adds `noise_scale: float` to the `BenchmarkSCM` Protocol. All four existing benchmarks do set `self.noise_scale` in their constructors, so this is safe. Good addition — it makes the implicit contract explicit.

**5. Threshold calculation: edge case with negative optimum**
`runner.py:244-250`

The threshold logic uses `abs(known_optimum)` as the scale factor. For a negative optimum (e.g., `-1.0`), the threshold is `0.10 * 1.0 = 0.10`, so a result of `-0.9` would be "within threshold" (`|-0.9 - (-1.0)| = 0.1 <= 0.1`). This is correct, but for very large negative optima (e.g., `-100.0`), the tolerance becomes `10.0`, which may be surprisingly generous. This is a reasonable design choice for benchmarks, just noting the behavior.

**6. `test_causal_beats_naive.py` — tolerance of `+1.0` is very generous**
`test_causal_beats_naive.py:56,93`

The assertion `avg_causal <= avg_random + 1.0` means the causal strategy could be *worse* than random by up to 1.0 and still pass. The docstrings explain the reasoning well (high variance with small budgets), but this is close to a no-op test. Consider whether you could tighten this over time (e.g., `avg_causal <= avg_random + 0.5`) or increase the budget/seeds to reduce variance.

Alternatively, consider a relative tolerance (e.g., causal must be within 20% of random's performance) rather than an absolute one, since the objective scale varies across benchmarks.

**7. `HighDimensionalSparseBenchmark.run()` ignores x4-x20 parameters silently**
`high_dimensional.py:69-80`

Parameters `x4` through `x20` are accepted but completely ignored. This is the correct causal behavior (they're not ancestors of the objective), and the docstring documents it. However, the `run()` method doesn't even read them — it would behave identically whether or not they're passed. This is fine and arguably the cleanest implementation, but worth noting that a caller can't distinguish "parameter was irrelevant" from "parameter was ignored due to a bug."

**8. Minor: `_sample_random_params` is a module-level function**
`runner.py:253-278`

This is a clean utility function, well-implemented with support for all `VariableType` variants. Good error messages for unsupported types and empty categorical choices. One minor note: for `INTEGER` type, the bounds are cast with `int()` which truncates (e.g., `int(2.7)` → `2`). This matches the existing convention in the codebase.

---

### Things Done Well

- **Excellent documentation**: Every class and method has clear docstrings explaining the design rationale, not just the API. The `test_causal_beats_naive.py` module docstring explaining why tolerances are generous is a model of good test documentation.
- **Proper RNG handling**: Using `SeedSequence.spawn()` for stream splitting in `_run_random` is the correct numpy pattern (even with the minor inconsistency noted above).
- **Protocol evolution**: Adding `noise_scale` to `BenchmarkSCM` is the right move — it makes the runner's `self.benchmark.noise_scale` access type-safe.
- **Test coverage is thorough**: Unit tests for the benchmark (space, graph, POMIS, determinism, irrelevant variables), smoke tests parameterized across all benchmarks × strategies, convergence curve monotonicity, edge cases (budget=1, threshold calculations), and the directional integration tests.
- **`test_known_pomis_matches_algorithm`**: Verifying hardcoded POMIS against the actual `compute_pomis()` algorithm is an excellent correctness check that catches drift.
- **Clean separation**: The runner doesn't embed strategy logic — it delegates to the engine or to simple random sampling, keeping the comparison framework orthogonal to the optimization logic.
- **`pytest.mark.slow`**: Good practice for gating the expensive directional tests. The `pyproject.toml` marker registration prevents warnings.

---

### Suggestions (non-blocking)

1. **Add a `summary()` or `__str__` method to `BenchmarkResult`** for quick inspection during development (e.g., printing strategy comparison tables).

2. **Consider adding `"surrogate_only"` to the directional tests** in `test_causal_beats_naive.py`. The current tests compare causal vs. random, but causal vs. surrogate-only would isolate the value of the causal graph specifically (vs. just having a surrogate model).

3. **The `compare()` method runs sequentially** — for larger benchmark suites, consider noting that parallelization (e.g., `concurrent.futures`) could be added later if needed. No need to implement now.

4. **`_run_random` accepts an `rng` parameter but constructs its own** via `SeedSequence` — simplify by removing the `rng` parameter and constructing both streams inside the method.

---

datablogin:
@greptileai

PR #15, opened 2026-03-12
Wire SensitivityValidator into ExperimentEngine

datablogin:
## Summary

- Wire `SensitivityValidator` into `ExperimentEngine` to validate improvements at each phase transition
- Add `validation_results: list[RobustnessReport]` to engine state, populated at exploration->optimization and optimization->exploitation transitions
- If `is_robust=False`, log a warning but do not block phase transitions (engine always makes forward progress)
- Add full pipeline integration test (`test_full_pipeline.py`): runs 60 experiments on ToyGraph with causal graph, verifies phase transitions, screening, reasonable results, and robustness reports
- Add no-graph pipeline integration test (`test_no_graph_pipeline.py`): runs 60 experiments without prior graph, verifies graph discovery produces non-empty graph and robustness reports are generated
- Add unit tests for validation wiring edge cases (few results, error handling)

## Test plan

- [x] Unit tests: `tests/unit/test_sensitivity_wiring.py` (6 tests)
  - Engine has `validation_results` attribute
  - RobustnessReports generated at phase transitions
  - Validation does not block phase transitions
  - Validation skipped with < 4 results
  - Graceful error handling when validator raises
- [x] Integration tests: `tests/integration/test_full_pipeline.py` (6 tests)
  - Phase transitions occur across 60 experiments
  - Screening runs and identifies variables
  - Off-policy predictor logging works
  - Final result is reasonable (objective < 0)
  - At least one RobustnessReport generated
  - Both KEEP and DISCARD statuses present
- [x] Integration tests: `tests/integration/test_no_graph_pipeline.py` (6 tests)
  - Auto-discovery produces non-empty graph
  - Active graph set from discovery
  - Full loop completes without errors
  - Phase transitions still happen
  - RobustnessReport generated
  - Reasonable result found
- [x] All 310 existing + new tests pass
- [x] Lint clean (ruff), format clean, mypy strict clean

<h3>Greptile Summary</h3>

This PR wires `SensitivityValidator` into `ExperimentEngine` so that a `RobustnessReport` is generated at each confirmed phase transition, adds a direction-aware override that downgrades regression reports to `is_robust=False`, and introduces a comprehensive suite of unit and integration tests. The placement of `_run_validation` after `_run_screening` ensures that reports are only emitted for transitions that actually stick, which correctly addresses the ordering concern raised in prior review.

**Key observations:**
- The validation placement after screening is correct and well-documented in both the code and the PR description.
- The direction-override logic (`improving = (effect_size < 0) if minimize else (effect_size > 0)`) properly handles the sign-blindness of `SensitivityValidator.abs(effect)`.
- `RobustnessReport` has no `old_phase`/`new_phase` fields — when screening retries generate multiple `exploration→optimization` reports, `validation_results` entries become ambiguous for any downstream consumer. This should be fixed before the feature is relied upon programmatically.
- `test_full_pipeline_off_policy_skips` asserts only that the logger produced output, which is guaranteed by phase transitions and validation logging and has nothing to do with off-policy skipping.

<h3>Confidence Score: 3/5</h3>

- Safe to merge with low risk, but the missing phase metadata in RobustnessReport should be addressed before downstream consumers rely on validation_results ordering.
- The core wiring logic is correct and well-tested. The two remaining issues are a missing metadata field on RobustnessReport (which will silently confuse consumers when screening retries occur) and a trivially-passing test assertion. Neither is a runtime failure, but the metadata gap is a real API design hole that will need to be papered over by any caller inspecting validation_results by index.
- causal_optimizer/engine/loop.py — the RobustnessReport construction at lines 717–728 should carry old_phase/new_phase; tests/integration/test_full_pipeline.py — test_full_pipeline_off_policy_skips needs a meaningful assertion.

| Filename | Overview |
|----------|----------|
| causal_optimizer/engine/loop.py | Wires SensitivityValidator into _update_phase; validation correctly fires after screening confirms the transition, and direction-override logic is sound. However, RobustnessReport lacks old_phase/new_phase metadata, making validation_results hard to interpret when screening retries generate multiple reports for the same conceptual transition. |
| tests/integration/test_full_pipeline.py | Good coverage overall, but test_full_pipeline_off_policy_skips asserts only that any log output was emitted — this trivially passes and does not verify off-policy skipping behaviour as its name/docstring claims. |
| tests/integration/test_no_graph_pipeline.py | Solid no-graph integration test covering discovery, phase transitions, and validation reports. The reasonable-result threshold (< 1.0) is intentionally relaxed relative to the full-graph test, which is appropriate. |
| tests/unit/test_sensitivity_wiring.py | Six targeted unit tests that address the issues raised in previous reviews (direction override, graceful error, skip-with-few-results). The assertions previously flagged as too weak or trivially passing have been corrected. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant EE as ExperimentEngine
    participant UP as _update_phase()
    participant SC as _run_screening()
    participant RV as _run_validation()
    participant SV as SensitivityValidator
    participant VR as validation_results[]

    EE->>UP: step() — n reaches threshold
    UP->>UP: old_phase = self._phase
    UP->>UP: set self._phase by n threshold
    alt old_phase=="exploration" and new phase in optimization/exploitation
        UP->>SC: _run_screening()
        SC-->>UP: may revert self._phase → "exploration"
    end
    alt old_phase != self._phase (transition confirmed)
        UP->>RV: _run_validation(old_phase, self._phase)
        RV->>RV: Guard: len(results) < 4 → return
        RV->>SV: validate_improvement(log, baseline_ids, improved_ids)
        SV-->>RV: RobustnessReport (effect_size, is_robust, …)
        alt is_robust AND effect direction wrong for minimize/maximize
            RV->>RV: Override: is_robust=False, rewrite summary
        end
        RV->>VR: append(report)
        RV->>EE: log info/warning
    end
```

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `causal_optimizer/engine/loop.py`, line 455-477 ([link](https://github.com/datablogin/causal-optimizer/blob/01a2a802ab19152be482fda93ba1e4da811920d4/causal_optimizer/engine/loop.py#L455-L477)) 

   **Validation fires before screening can revert the phase**

   `_run_validation` is called before `_run_screening`, but `_run_screening` can set `self._phase = "exploration"` when no important variables are found. This means:

   1. A `RobustnessReport` claiming an `"exploration→optimization"` transition is appended to `validation_results`.
   2. Screening then reverts the phase back to `"exploration"`.
   3. On the very next `step()` call the same tentative transition occurs again (`n ≥ 10`, `old_phase == "exploration"`, `self._phase` reset to `"optimization"`), triggering `_run_validation` **again** with nearly identical data.

   This produces a series of spurious, duplicate reports that each claim a transition that never completed, inflating `validation_results` and misleading callers who inspect those reports.

   The fix is to move the validation check **after** the screening / discovery block so it only fires when the final post-screening phase differs from `old_phase`:

   ```python
   # Run screening and POMIS when transitioning from exploration to optimization
   if old_phase == "exploration" and self._phase in ("optimization", "exploitation"):
       self._run_auto_discovery()

       if self._screening_attempts < self._max_screening_attempts:
           self._run_screening()
       else:
           logger.warning(
               "Max screening attempts (%d) reached with no important variables; "
               "proceeding with all variables",
               self._max_screening_attempts,
           )
           self._screened_focus_variables = self.search_space.variable_names
       if self._phase == "optimization":
           self._compute_pomis()

   # Validate AFTER screening has had its final say on the phase
   if old_phase != self._phase:
       self._run_validation(old_phase, self._phase)
   ```

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   Path: causal_optimizer/engine/loop.py
   Line: 455-477

   Comment:
   **Validation fires before screening can revert the phase**

   `_run_validation` is called before `_run_screening`, but `_run_screening` can set `self._phase = "exploration"` when no important variables are found. This means:

   1. A `RobustnessReport` claiming an `"exploration→optimization"` transition is appended to `validation_results`.
   2. Screening then reverts the phase back to `"exploration"`.
   3. On the very next `step()` call the same tentative transition occurs again (`n ≥ 10`, `old_phase == "exploration"`, `self._phase` reset to `"optimization"`), triggering `_run_validation` **again** with nearly identical data.

   This produces a series of spurious, duplicate reports that each claim a transition that never completed, inflating `validation_results` and misleading callers who inspect those reports.

   The fix is to move the validation check **after** the screening / discovery block so it only fires when the final post-screening phase differs from `old_phase`:

   ```python
   # Run screening and POMIS when transitioning from exploration to optimization
   if old_phase == "exploration" and self._phase in ("optimization", "exploitation"):
       self._run_auto_discovery()

       if self._screening_attempts < self._max_screening_attempts:
           self._run_screening()
       else:
           logger.warning(
               "Max screening attempts (%d) reached with no important variables; "
               "proceeding with all variables",
               self._max_screening_attempts,
           )
           self._screened_focus_variables = self.search_space.variable_names
       if self._phase == "optimization":
           self._compute_pomis()

   # Validate AFTER screening has had its final say on the phase
   if old_phase != self._phase:
       self._run_validation(old_phase, self._phase)
   ```

   `````
   </details>

   <a href="https://app.greptile.com/ide/claude-code?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20causal_optimizer%2Fengine%2Floop.py%0ALine%3A%20455-477%0A%0AComment%3A%0A**Validation%20fires%20before%20screening%20can%20revert%20the%20phase**%0A%0A%60_run_validation%60%20is%20called%20before%20%60_run_screening%60%2C%20but%20%60_run_screening%60%20can%20set%20%60self._phase%20%3D%20%22exploration%22%60%20when%20no%20important%20variables%20are%20found.%20This%20means%3A%0A%0A1.%20A%20%60RobustnessReport%60%20claiming%20an%20%60%22exploration%E2%86%92optimization%22%60%20transition%20is%20appended%20to%20%60validation_results%60.%0A2.%20Screening%20then%20reverts%20the%20phase%20back%20to%20%60%22exploration%22%60.%0A3.%20On%20the%20very%20next%20%60step%28%29%60%20call%20the%20same%20tentative%20transition%20occurs%20again%20%28%60n%20%E2%89%A5%2010%60%2C%20%60old_phase%20%3D%3D%20%22exploration%22%60%2C%20%60self._phase%60%20reset%20to%20%60%22optimization%22%60%29%2C%20triggering%20%60_run_validation%60%20**again**%20with%20nearly%20identical%20data.%0A%0AThis%20produces%20a%20series%20of%20spurious%2C%20duplicate%20reports%20that%20each%20claim%20a%20transition%20that%20never%20completed%2C%20inflating%20%60validation_results%60%20and%20misleading%20callers%20who%20inspect%20those%20reports.%0A%0AThe%20fix%20is%20to%20move%20the%20validation%20check%20**after**%20the%20screening%20%2F%20discovery%20block%20so%20it%20only%20fires%20when%20the%20final%20post-screening%20phase%20differs%20from%20%60old_phase%60%3A%0A%0A%60%60%60python%0A%23%20Run%20screening%20and%20POMIS%20when%20transitioning%20from%20exploration%20to%20optimization%0Aif%20old_phase%20%3D%3D%20%22exploration%22%20and%20self._phase%20in%20%28%22optimization%22%2C%20%22exploitation%22%29%3A%0A%20%20%20%20self._run_auto_discovery%28%29%0A%0A%20%20%20%20if%20self._screening_attempts%20%3C%20self._max_screening_attempts%3A%0A%20%20%20%20%20%20%20%20self._run_screening%28%29%0A%20%20%20%20else%3A%0A%20%20%20%20%20%20%20%20logger.warning%28%0A%20%20%20%20%20%20%20%20%20%20%20%20%22Max%20screening%20attempts%20%28%25d%29%20reached%20with%20no%20important%20variables%3B%20%22%0A%20%20%20%20%20%20%20%20%20%20%20%20%22proceeding%20with%20all%20variables%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20self._max_screening_attempts%2C%0A%20%20%20%20%20%20%20%20%29%0A%20%20%20%20%20%20%20%20self._screened_focus_variables%20%3D%20self.search_space.variable_names%0A%20%20%20%20if%20self._phase%20%3D%3D%20%22optimization%22%3A%0A%20%20%20%20%20%20%20%20self._compute_pomis%28%29%0A%0A%23%20Validate%20AFTER%20screening%20has%20had%20its%20final%20say%20on%20the%20phase%0Aif%20old_phase%20!%3D%20self._phase%3A%0A%20%20%20%20self._run_validation%28old_phase%2C%20self._phase%29%0A%60%60%60%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&repo=datablogin%2Fcausal-optimizer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/badge/Fix_in_Claude-black?logo=claude&logoColor=%23D97706">Fix in Claude Code</picture></a>
</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%202%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%202%0Acausal_optimizer%2Fengine%2Floop.py%3A717-730%0A**%60RobustnessReport%60%20carries%20no%20phase-transition%20metadata**%0A%0A%60validation_results%60%20can%20accumulate%20more%20than%20two%20reports%20when%20screening%20retries%20cause%20repeated%20%60exploration%E2%86%92optimization%60%20transitions.%20Each%20retry%20calls%20%60_run_validation%60%20and%20appends%20a%20fresh%20%60RobustnessReport%60%2C%20but%20%60RobustnessReport%60%20has%20no%20%60old_phase%60%2F%60new_phase%60%20fields.%20A%20downstream%20consumer%20iterating%20%60engine.validation_results%60%20therefore%20cannot%20determine%20which%20transition%20any%20given%20report%20corresponds%20to%20%E2%80%94%20%60results%5B0%5D%60%20might%20be%20the%20first%20screening-confirmed%20%60exploration%E2%86%92optimization%60%2C%20%60results%5B1%5D%60%20might%20be%20a%20retry%20of%20the%20same%20transition%2C%20and%20%60results%5B2%5D%60%20the%20%60optimization%E2%86%92exploitation%60%20transition.%20Positional%20indexing%20is%20fragile%20and%20will%20break%20silently%20whenever%20screening%20retries%20change%20the%20count.%0A%0AThe%20fix%20is%20either%20to%20extend%20%60RobustnessReport%60%20with%20optional%20phase%20fields%2C%20or%20to%20introduce%20a%20thin%20wrapper%3A%0A%0A%60%60%60python%0A%23%20In%20sensitivity.py%0A%40dataclass%0Aclass%20RobustnessReport%3A%0A%20%20%20%20effect_size%3A%20float%0A%20%20%20%20noise_estimate%3A%20float%0A%20%20%20%20signal_to_noise%3A%20float%0A%20%20%20%20e_value%3A%20float%0A%20%20%20%20is_robust%3A%20bool%0A%20%20%20%20summary%3A%20str%0A%20%20%20%20old_phase%3A%20str%20%3D%20%22%22%20%20%20%23%20phase%20before%20the%20transition%0A%20%20%20%20new_phase%3A%20str%20%3D%20%22%22%20%20%20%23%20phase%20after%20the%20transition%0A%60%60%60%0A%0Aand%20populate%20them%20in%20%60_run_validation%60%3A%0A%0A%60%60%60python%0Areport%20%3D%20RobustnessReport%28%0A%20%20%20%20...%2C%0A%20%20%20%20is_robust%3DFalse%2C%0A%20%20%20%20summary%3D...%2C%0A%20%20%20%20old_phase%3Dold_phase%2C%0A%20%20%20%20new_phase%3Dnew_phase%2C%0A%29%0A%60%60%60%0A%0A%23%23%23%20Issue%202%20of%202%0Atests%2Fintegration%2Ftest_full_pipeline.py%3A62-72%0A**Test%20assertion%20doesn't%20verify%20its%20stated%20purpose**%0A%0AThe%20test%20is%20named%20%60test_full_pipeline_off_policy_skips%60%20and%20the%20docstring%20states%20*%22Off-policy%20predictor%20should%20skip%20at%20least%20some%20experiments%22*%2C%20but%20the%20only%20assertion%20is%3A%0A%0A%60%60%60python%0Aassert%20len%28caplog.records%29%20%3E%200%2C%20%22Expected%20log%20output%20from%20the%20engine%20loop%22%0A%60%60%60%0A%0AThis%20is%20trivially%20true%20%E2%80%94%20the%20engine%20emits%20multiple%20log%20messages%20during%20any%2060-experiment%20run%20%28phase%20transitions%2C%20screening%2C%20validation%29%2C%20so%20the%20assertion%20passes%20regardless%20of%20whether%20the%20off-policy%20predictor%20ever%20skipped%20anything.%20It%20doesn't%20actually%20test%20the%20claimed%20behaviour.%0A%0ATo%20make%20the%20test%20meaningful%2C%20verify%20that%20at%20least%20one%20%22skip%22%20log%20record%20was%20emitted%2C%20e.g.%3A%0A%0A%60%60%60python%0Askip_records%20%3D%20%5Br%20for%20r%20in%20caplog.records%20if%20%22skip%22%20in%20r.getMessage%28%29.lower%28%29%5D%0Aassert%20len%28skip_records%29%20%3E%200%2C%20%28%0A%20%20%20%20%22Expected%20off-policy%20predictor%20to%20skip%20at%20least%20one%20experiment%20in%2060%20runs%22%0A%29%0A%60%60%60%0A%0AOr%2C%20if%20skipping%20is%20genuinely%20non-deterministic%20for%20this%20seed%2Fbenchmark%2C%20either%20adjust%20the%20fixture%20to%20a%20scenario%20that%20guarantees%20skips%20or%20rename%20the%20test%20to%20reflect%20what%20it%20actually%20checks%20%28e.g.%2C%20%60test_full_pipeline_produces_log_output%60%29.%0A%0A&repo=datablogin%2Fcausal-optimizer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/badge/Fix_All_in_Claude-black?logo=claude&logoColor=%23D97706">Fix All in Claude Code</picture></a>

`````markdown
Path: causal_optimizer/engine/loop.py
Line: 717-730

Comment:
**`RobustnessReport` carries no phase-transition metadata**

`validation_results` can accumulate more than two reports when screening retries cause repeated `exploration→optimization` transitions. Each retry calls `_run_validation` and appends a fresh `RobustnessReport`, but `RobustnessReport` has no `old_phase`/`new_phase` fields. A downstream consumer iterating `engine.validation_results` therefore cannot determine which transition any given report corresponds to — `results[0]` might be the first screening-confirmed `exploration→optimization`, `results[1]` might be a retry of the same transition, and `results[2]` the `optimization→exploitation` transition. Positional indexing is fragile and will break silently whenever screening retries change the count.

The fix is either to extend `RobustnessReport` with optional phase fields, or to introduce a thin wrapper:

```python
# In sensitivity.py
@dataclass
class RobustnessReport:
    effect_size: float
    noise_estimate: float
    signal_to_noise: float
    e_value: float
    is_robust: bool
    summary: str
    old_phase: str = ""   # phase before the transition
    new_phase: str = ""   # phase after the transition
```

and populate them in `_run_validation`:

```python
report = RobustnessReport(
    ...,
    is_robust=False,
    summary=...,
    old_phase=old_phase,
    new_phase=new_phase,
)
```

---

Path: tests/integration/test_full_pipeline.py
Line: 62-72

Comment:
**Test assertion doesn't verify its stated purpose**

The test is named `test_full_pipeline_off_policy_skips` and the docstring states *"Off-policy predictor should skip at least some experiments"*, but the only assertion is:

```python
assert len(caplog.records) > 0, "Expected log output from the engine loop"
```

This is trivially true — the engine emits multiple log messages during any 60-experiment run (phase transitions, screening, validation), so the assertion passes regardless of whether the off-policy predictor ever skipped anything. It doesn't actually test the claimed behaviour.

To make the test meaningful, verify that at least one "skip" log record was emitted, e.g.:

```python
skip_records = [r for r in caplog.records if "skip" in r.getMessage().lower()]
assert len(skip_records) > 0, (
    "Expected off-policy predictor to skip at least one experiment in 60 runs"
)
```

Or, if skipping is genuinely non-deterministic for this seed/benchmark, either adjust the fixture to a scenario that guarantees skips or rename the test to reflect what it actually checks (e.g., `test_full_pipeline_produces_log_output`).

`````

</details>

<sub>Last reviewed commit: e709fa0</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/tests/unit/test_sensitivity_wiring.py
+++ b/tests/unit/test_sensitivity_wiring.py
@@ -163,3 +163,32 @@ def test_validation_wrong_direction_treated_as_non_robust() -> None:
     assert report.is_robust is False
     assert "Effect direction is wrong" in report.summary
     assert "originally:" in report.summary
+
+
+def test_validation_wrong_direction_maximize() -> None:
+    """Direction override also works for maximize: negative effect is wrong."""
+    engine = ExperimentEngine(
+        search_space=_make_space(),
+        runner=_QuadRunner(),
+        minimize=False,
+        seed=42,
+    )
+    for _ in range(11):
+        engine.run_experiment({"x": 1.0, "y": 1.0})
+
+    # effect_size < 0 is wrong for maximization
+    mock_report = RobustnessReport(
+        effect_size=-2.0,
+        noise_estimate=0.5,
+        signal_to_noise=4.0,
+        e_value=6.0,
+        is_robust=True,
+        summary="Robust improvement",
+    )
+    with patch.object(engine._validator, "validate_improvement", return_value=mock_report):
+        engine._run_validation("exploration", "optimization")
+
+    assert len(engine.validation_results) == 1
+    report = engine.validation_results[0]
+    assert report.is_robust is False
+    assert "Effect direction is wrong" in report.summary

### CODE REVIEW

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the PR. Here's my review:

---

## PR Review: Wire SensitivityValidator into ExperimentEngine

### Overall Assessment

This is a clean, well-structured integration PR. The wiring approach is sound — validation runs after confirmed phase transitions, is non-blocking, and has proper error handling. Good test coverage across unit and integration layers. A few issues worth addressing:

---

### Issues

**1. Validation compares arbitrary halves, not phase-specific experiments (Design concern)**

`_run_validation` at `loop.py:687-689` splits *all* experiments at the midpoint:

```python
midpoint = len(results) // 2
baseline_ids = [r.experiment_id for r in results[:midpoint]]
improved_ids = [r.experiment_id for r in results[midpoint:]]
```

At the exploration→optimization transition (n=10), this compares experiments 1-5 vs 6-10. That's reasonable for a within-phase trend check. But at the optimization→exploitation transition (n=50), it compares experiments 1-25 vs 26-50, which spans the exploration→optimization boundary. The "baseline" includes exploration experiments (random LHS) and the "improved" includes optimization experiments — you're comparing apples to oranges since the strategies are fundamentally different.

Consider filtering to only experiments from `old_phase` (using the metadata `"phase"` key), or at least document this as an intentional design choice. The docstring says "within-phase trend check" but the implementation is cross-phase.

**2. `effect_size == 0` edge case in direction check**

`loop.py:710`:
```python
improving = (report.effect_size < 0) if self.minimize else (report.effect_size > 0)
```

When `effect_size == 0.0`, `improving` is `False`. If the report also has `is_robust=True` (unlikely given SNR would be 0, but possible with floating point), the direction override triggers and rewrites the report. This is probably fine in practice since `is_robust` requires `snr > 1.0`, but worth a brief comment noting the `== 0` case is intentionally treated as "not improving."

**3. `RobustnessReport` is not frozen**

Per `sensitivity.py:22-31`, `RobustnessReport` is a mutable dataclass. In `_run_validation`, instead of mutating the original report, you construct a new one (good). But consumers could accidentally mutate reports in `validation_results`. Consider `@dataclass(frozen=True)` — it's a pure value object with no mutation use case.

*Caveat:* Check whether other code mutates it. Since this PR introduces the wiring, and the only place `RobustnessReport` is created is inside `SensitivityValidator.validate_improvement` and `_run_validation`, freezing should be safe.

**4. Integration tests each run 60 experiments independently — slow test suite**

Both integration test files create fresh engines and run 60-step loops per test function. With 6 tests in `test_full_pipeline.py` and 6 in `test_no_graph_pipeline.py`, that's 12 × 60 = 720 engine steps. Each step involves surrogate fitting, screening, etc.

Consider either:
- Using a session-scoped fixture that runs the loop once and shares the result across tests
- Or marking these as `@pytest.mark.slow` so they can be skipped in fast CI runs

This is a pragmatic concern — if your CI is already fast, ignore this.

**5. `validation_transitions` parallel list pattern is fragile**

`loop.py:179-181`:
```python
self.validation_results: list[RobustnessReport] = []
self.validation_transitions: list[tuple[str, str]] = []
```

Parallel lists that must stay in sync are a classic maintenance hazard. A more robust approach would be a small dataclass:

```python
@dataclass
class ValidationRecord:
    report: RobustnessReport
    old_phase: str
    new_phase: str
```

Then `self.validation_records: list[ValidationRecord] = []`. This eliminates the sync invariant and is more self-documenting for consumers.

**6. `SensitivityValidator` is unconditionally instantiated**

`loop.py:176`:
```python
self._validator = SensitivityValidator()
```

This is always created even if no phase transitions will ever occur (e.g., `run_loop(n_experiments=5)`). It's a lightweight object so the cost is negligible, but it introduces a hard import dependency on `scipy.stats` at engine construction time. If scipy is an optional dependency in some configurations, this could break. Check whether `scipy` is in the core or optional dependency group.

---

### Minor Nits

- `loop.py:683` — The `< 4` guard is good, but the comment says "2 baseline + 2 improved." Since `SensitivityValidator.validate_improvement` also has a `< 2` guard per group (`sensitivity.py:59`), the engine-side check is redundant but harmless. Consider removing the engine-side check and relying on the validator, or document the intentional double-guard.

- `test_sensitivity_wiring.py:77-80` — `test_validation_does_not_block_phase_transition` checks that `"optimization" in phases_seen or "exploitation" in phases_seen`, but since screening can revert to exploration, this test could flake if screening happens to revert all attempts. The assertion message correctly calls this out, but the test is checking a negative (validation doesn't block) via an indirect observation (phase advanced). A more direct test would mock `_run_validation` to raise and confirm the phase still advances.

- `test_full_pipeline.py:108` — `assert hasattr(engine_with_graph, "validation_results")` is unnecessary since the attribute is set in `__init__`. If it were missing, the preceding `run_loop` call would have crashed. This check adds no value.

---

### What's Done Well

- **Non-blocking design**: Validation warns but never blocks forward progress. This is the right call for an optimization engine.
- **Direction override logic**: Catching "robust regression" (statistically significant worsening) and relabeling it as non-robust is a subtle and correct design decision.
- **Graceful error handling**: The bare `except Exception` with `exc_info=True` logging is appropriate for a validation side-channel that must not crash the engine.
- **Test coverage**: Good spread across unit tests (mocking, edge cases, direction override) and integration tests (full pipeline with/without graph). The `_QuadRunner` test helper is clean and minimal.
- **Comments**: The inline comments explaining *why* validation runs after screening (not before) are helpful.

---

### Summary

The core wiring logic is correct and well-integrated. The main actionable items are (1) the cross-phase comparison semantics at the opt→exploit boundary, (2) replacing parallel lists with a proper record type, and (3) considering frozen dataclass for `RobustnessReport`. Everything else is minor.

---

datablogin:
# Claude Code Review

## Review Feedback

Here's my review of this PR:

---

## PR Review: Wire SensitivityValidator into ExperimentEngine

**Overall assessment:** This is a clean, well-structured integration. The wiring is non-blocking (validation never prevents forward progress), the error handling is defensive, and the direction-check logic is a good safety net. The tests are thorough. I have a few issues to raise, mostly around a subtle semantic problem and minor improvements.

---

### Issues

**1. Baseline/improved split semantics are misleading (medium)**

`_run_validation` at `loop.py:703-705` splits experiments 50/50 by index order and labels them "baseline" vs "improved":

```python
midpoint = len(results) // 2
baseline_ids = [r.experiment_id for r in results[:midpoint]]
improved_ids = [r.experiment_id for r in results[midpoint:]]
```

This assumes later experiments are better than earlier ones, which is only true if the optimizer is making progress. But at the exploration→optimization boundary, the first 10 experiments are space-filling LHS — there's no reason the second half should be "improved" relative to the first half. The labels `baseline_experiments` / `improved_experiments` in `SensitivityValidator.validate_improvement` carry strong semantic meaning (the method name itself says "improvement"), but the engine is just splitting chronologically.

The practical impact is limited because validation doesn't block anything, but it means the `is_robust` flag and the direction-override logic can produce confusing results. Consider:
- Documenting that this is a "trajectory trend check" rather than a true baseline-vs-treatment comparison
- Or splitting by phase (all exploration-phase experiments as baseline, vs. the last N optimization experiments)

The docstring acknowledges this somewhat ("within-phase trend check"), but the variable names and the call into `validate_improvement` still suggest a treatment-vs-control design.

**2. `RobustnessReport` made `frozen=True` but reconstructed via field spread (low)**

In `sensitivity.py:22`, changing `@dataclass` to `@dataclass(frozen=True)` is good for immutability. But the override path in `_run_validation` (`loop.py:738-749`) reconstructs the report by manually copying every field. If `RobustnessReport` gains a new field in the future, this reconstruction will silently drop it, producing a report with default values for the new field (or raising a `TypeError` if the new field has no default).

Consider using `dataclasses.replace()` instead:

```python
from dataclasses import replace
report = replace(report, is_robust=False, summary=f"Effect direction is wrong ...")
```

This is safer against schema drift and is the idiomatic pattern for frozen dataclasses.

**3. `validation_results` property creates a new list on every call (low)**

`loop.py:193-194`:
```python
def validation_results(self) -> list[RobustnessReport]:
    return [r.report for r in self.validation_records]
```

This allocates a new list every call. Fine for typical usage, but if someone calls `engine.validation_results` in a loop or hot path, it's wasteful. Not a blocker, just worth noting — the docstring could mention this returns a fresh list, or it could be cached.

**4. Integration tests re-run the full 60-experiment loop per test (low)**

Each test in `test_full_pipeline.py` and `test_no_graph_pipeline.py` runs `engine.run_loop(n_experiments=60)` independently. That's 6 full pipeline runs in `test_full_pipeline.py` alone. Since these are marked `@pytest.mark.slow`, this is intentional. But you could consolidate with a session-scoped or module-scoped fixture that runs once and shares the result across tests, cutting test time significantly. The tests are all read-only assertions on the final state.

The counter-argument is test isolation, which is fair — just flagging the tradeoff.

**5. `test_validation_does_not_block_phase_transition` has a weak assertion (low)**

`test_sensitivity_wiring.py:79-82`:
```python
phases_seen = {r.metadata.get("phase") for r in engine.log.results}
assert "optimization" in phases_seen or "exploitation" in phases_seen
```

This doesn't actually prove validation didn't block the transition — it only proves the phase advanced, which would happen regardless. To truly test this, you'd need to mock `_validator.validate_improvement` to return `is_robust=False` and then verify the phase still advances. As written, the test passes even without any validation wiring at all.

**6. Broad `except Exception` in `_run_validation` (nit)**

`loop.py:714` catches all exceptions from `validate_improvement`. This is a reasonable choice for a non-critical validation step that should never crash the engine. But it does mean programming errors (e.g., `TypeError` from a bad argument, `AttributeError` from a refactor) will be silently logged and swallowed. Consider catching a narrower set like `(ValueError, RuntimeError, ArithmeticError)` — similar to what `_run_auto_discovery` does on line 542.

---

### Things done well

- **Non-blocking design:** Validation never prevents phase transitions. This is exactly right for an informational check.
- **Direction-override logic:** Catching the case where `SensitivityValidator` says "robust" but the effect goes the wrong way is a smart guard. The `abs(effect)` used by SNR/E-value makes this a real concern.
- **Frozen dataclass for `RobustnessReport`:** Good change — reports should be immutable once created.
- **`ValidationRecord` bundling phase context:** Clean design. Having `old_phase` and `new_phase` alongside the report makes records self-describing.
- **Test coverage:** The unit tests cover the happy path, error path, insufficient data, and wrong-direction cases well. The mock-based tests for error handling and direction override are well-constructed.
- **Comments and docstrings:** Consistently explain *why*, not just *what*. The comment at line 491-493 explaining why validation runs after screening is particularly helpful.

---

PR #20, opened 2026-03-14
feat: SQLite persistence, engine resume, and CLI

datablogin:
## Summary

- **SQLite persistence** (`causal_optimizer/storage/sqlite.py`): Thread-safe `ExperimentStore` with lossless `ExperimentResult` round-trips via Pydantic `model_dump`/`model_validate`. Pure stdlib (`sqlite3`, `json`, `pathlib`). Context manager support, WAL journal mode, foreign key enforcement.
- **Engine integration** (`engine/loop.py`): Added `store`/`experiment_id` params to `ExperimentEngine.__init__()`, auto-persist after each `step()`, and `resume()` classmethod that reconstructs phase from result count. Validates store/experiment_id are paired. Restores MAP-Elites archive on resume.
- **CLI** (`causal_optimizer/cli.py` + `__main__.py`): `argparse`-based CLI with `run`, `resume`, `report`, and `list` subcommands. Adapter loading via `module:ClassName` spec. Entry point registered in `pyproject.toml`. Duplicate experiment ID detection, proper error handling for missing experiments.
- **15 new tests**: 6 store CRUD tests, 5 engine persistence integration tests (including phase inference and validation), 4 CLI subprocess tests.

## Test plan

- [x] `uv run pytest tests/unit/test_store.py tests/unit/test_cli.py -v` -- all 15 new tests pass
- [x] `uv run pytest -m "not slow"` -- all 321 tests pass (no regressions)
- [x] `uv run mypy causal_optimizer/` -- no type errors
- [x] `uv run ruff check .` -- no lint errors
- [x] `uv run causal-optimizer --help` -- exits 0

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/causal_optimizer/storage/sqlite.py
+++ b/causal_optimizer/storage/sqlite.py
@@ -58,6 +58,9 @@ class ExperimentStore:
             # Set after executescript because it issues an implicit COMMIT
             # that can reset connection-level pragmas.
             self._conn.execute("PRAGMA foreign_keys = ON")
+            # WAL mode improves concurrent read performance and reduces
+            # write contention when multiple processes access the same DB.
+            self._conn.execute("PRAGMA journal_mode = WAL")
 
     def experiment_exists(self, experiment_id: str) -> bool:
         """Check if an experiment with the given ID exists."""
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 from typing import Any
 
+import pytest
+
 from causal_optimizer.engine.loop import ExperimentEngine
 from causal_optimizer.storage.sqlite import ExperimentStore
 from causal_optimizer.types import (
@@ -198,6 +200,27 @@ def test_engine_resume_continues_from_step() -> None:
     assert len(log.results) == 10
     # Engine2 should also have 10 results in its in-memory log
     assert len(engine2.log.results) == 10
+
+
+def test_engine_rejects_store_without_experiment_id() -> None:
+    """Passing store without experiment_id raises ValueError."""
+    store = ExperimentStore(":memory:")
+    with pytest.raises(ValueError, match="store and experiment_id must be provided together"):
+        ExperimentEngine(
+            search_space=_make_search_space(),
+            runner=QuadraticRunner(),
+            store=store,
+        )
+
+
+def test_engine_rejects_experiment_id_without_store() -> None:
+    """Passing experiment_id without store raises ValueError."""
+    with pytest.raises(ValueError, match="store and experiment_id must be provided together"):
+        ExperimentEngine(
+            search_space=_make_search_space(),
+            runner=QuadraticRunner(),
+            experiment_id="test-id",
+        )
 
 
 def test_engine_resume_infers_optimization_phase() -> None:


datablogin:
@greptileai

datablogin:
@greptileai

datablogin:
# Claude Code Review

## Review Feedback

Here's my review of this PR:

---

## PR Review: SQLite Persistence, Engine Resume, and CLI

Overall this is a well-structured PR that adds a clean persistence layer and CLI to the engine. The code follows established project patterns and the test coverage hits the main paths. Below are specific issues and suggestions.

### Bugs / Correctness Issues

**1. `resume()` calls `create_experiment()` which silently overwrites the search space (Medium)**

`loop.py:225-230` — `resume()` calls `cls(...)` which triggers `__init__`, which calls `store.create_experiment()` with `INSERT OR IGNORE`. This means on resume, the constructor attempts to re-create the experiment. While `INSERT OR IGNORE` makes this a no-op, it's semantically confusing — and if the schema ever changes to `INSERT OR REPLACE`, it would silently overwrite the stored search space. Consider either:
- Skipping the `create_experiment` call when resuming (e.g., add an internal flag)
- Or having `resume()` set `store`/`experiment_id` after construction to avoid the duplicate create

**2. `resume()` does not restore screening state, validation records, or POMIS sets (Medium)**

`loop.py:232-256` — Phase is inferred from result count, but `_screening_result`, `_screened_focus_variables`, `_pomis_sets`, and `validation_records` are not restored. If an experiment was interrupted mid-optimization phase (e.g., at step 30), the resumed engine will be in "optimization" phase but without the screening results that originally guided it. This means:
- `suggest_next()` won't pass `screened_variables` to `suggest_parameters()`
- POMIS-guided suggestions won't work
- The next `_update_phase()` call won't re-trigger screening (it only fires on phase *transitions*)

This could silently degrade optimization quality after resume. At minimum, document this limitation. Ideally, re-run screening on resume when in optimization phase.

**3. `resume()` doesn't restore `_screening_attempts` counter**

`loop.py:232-256` — If screening was attempted but failed (setting `_screening_attempts > 0`) and caused a phase revert, a resumed engine resets to `_screening_attempts = 0`, potentially causing an infinite revert loop if the data consistently produces no important variables.

**4. `list_experiments` COUNT may be wrong with `results.id` (Low)**

`sqlite.py:144` — `COUNT(r.id)` counts non-null `r.id` values. Since `results.id` stores the `ExperimentResult.experiment_id` (which is always non-null for inserted rows), this works correctly — but `COUNT(r.step)` or `COUNT(*)` in the joined rows would be more semantically clear for counting result rows.

### Thread Safety

**5. PRAGMA `foreign_keys` is not protected by the lock**

`sqlite.py:36` — `_create_tables` acquires `self._lock`, but the `PRAGMA foreign_keys = ON` call at line 36 is outside the lock. While this is only called during `__init__` (so no contention yet), it breaks the pattern established elsewhere.

**6. Lock granularity allows read-your-writes issues**

`sqlite.py:112-138` — In `load_log()`, the lock is held for the SELECT but released before the deserialization loop. This is fine since SQLite serializes writes at the DB level, but worth noting that the lock pattern here is primarily for Python-level thread safety of the `_conn` object, not transactional isolation.

### Design / API

**7. `results.id` column naming is confusing**

`sqlite.py:48-49` — The `results` table has an `id` column that stores `ExperimentResult.experiment_id` (the per-result UUID), while `experiment_id` is the foreign key to the experiments table. This naming collision (`results.id` is NOT the primary key, and `results.experiment_id` != `results.id`) will confuse future readers. Consider renaming to `result_id` for clarity.

**8. No primary key on results table**

`sqlite.py:47-58` — The `results` table has `UNIQUE(experiment_id, step)` but no explicit primary key. SQLite will auto-create a `rowid`, which is fine, but an explicit `PRIMARY KEY(experiment_id, step)` would be more intentional.

**9. `ExperimentStore` not closed in CLI tests**

`test_cli.py:60-61, 156-157` — `ExperimentStore` is created for verification but never closed. Should use a context manager:
```python
with ExperimentStore(db_path) as store:
    ...
```

**10. `_load_adapter` instantiates with no arguments**

`cli.py:45` — `cls()` assumes the adapter's `__init__` takes no arguments. This works for the toy adapter but will fail for any adapter requiring configuration (e.g., API keys, connection strings). Consider supporting a `--adapter-config` JSON file or documenting this limitation.

### Missing Test Coverage

**11. No test for `resume()` in optimization/exploitation phases**

The resume test only exercises the exploration phase (3+3=6 steps, all under the `_EXPLORATION_LIMIT=10` threshold). There's no test for resuming an experiment that had reached optimization (10-49 steps) or exploitation (50+ steps). These are the phases where state restoration matters most (screening, POMIS, MAP-Elites).

**12. No test for the `--seed` flag**

The CLI `run` parser accepts `--seed` but no test verifies reproducibility.

**13. No test for error paths in CLI**

Missing tests for: invalid adapter spec, non-existent DB for report/list, duplicate experiment ID on `run`, resume of non-existent experiment.

**14. `test_cli_list_empty_db` uses `del store` instead of `store.close()`**

`test_cli.py:28` — `del store` relies on `__del__` for cleanup, which is not guaranteed to run immediately (or at all). Use `store.close()` explicitly.

### Minor Issues

**15. `store` and `experiment_id` must be passed together but this isn't validated**

`loop.py:203` — If `store` is provided but `experiment_id` is None (or vice versa), the engine silently ignores persistence. Consider raising `ValueError` if exactly one is provided.

**16. `_cmd_run` accesses `engine.log` after the context manager closes the store**

`cli.py:86-89` — `engine.log` is an in-memory object so this works, but the `with` block's intent suggests the store should be open during result access. Minor style concern.

**17. `__main__.py` doesn't guard with `if __name__ == "__main__"`**

`__main__.py:5` — While `__main__.py` is only executed as a module entry point, adding the guard is a defensive best practice.

### What's Done Well

- Clean separation between `ExperimentStore` (persistence) and `ExperimentEngine` (logic)
- `INSERT OR IGNORE` for idempotent `create_experiment` calls
- Thread safety via `threading.Lock` (appropriate for SQLite's threading model)
- Context manager protocol on `ExperimentStore`
- The `_AdapterRunner` pattern cleanly bridges `DomainAdapter` and `ExperimentRunner` protocols
- Phase constant extraction (`_EXPLORATION_LIMIT`, `_OPTIMIZATION_LIMIT`) eliminates magic numbers
- Good roundtrip test for `ExperimentResult` serialization
- CLI subprocess tests verify end-to-end behavior

---

PR #23, opened 2026-03-14
CLI feature parity: adapter config hooks and override flags

datablogin:
## Summary

- Extend `DomainAdapter` with 6 optional configuration hooks (`get_objective_name`, `get_minimize`, `get_strategy`, `get_objectives`, `get_constraints`, `get_discovery_method`) with backward-compatible defaults
- Add CLI flags to `run` and `resume` subcommands: `--objective-name`, `--minimize`/`--maximize`, `--strategy`, `--discovery-method`
- Add `--objective-name` and `--minimize`/`--maximize` flags to `report` subcommand
- Implement priority chain: CLI flag > adapter method > engine default
- Fix hardcoded `'objective'` key in `_cmd_run()`, `_cmd_resume()`, and `_cmd_report()` to use actual objective name
- Fix `best_result()` calls to pass configured `objective_name` and `minimize` values
- Wire `_adapter_engine_kwargs()` to pull config from all adapter hooks

## Test plan

- [x] 17 tests in `tests/unit/test_cli_extended.py` (all passing)
- [x] All 354+ existing fast tests still pass
- [x] `ruff check` clean
- [x] `ruff format` clean
- [x] `mypy` strict mode clean
- [x] Greptile review: 0 unresolved comments
- [x] Claude review: 0 actionable issues remaining

### DIFF
--- a/tests/unit/test_cli_extended.py
+++ b/tests/unit/test_cli_extended.py
@@ -11,8 +11,10 @@ from __future__ import annotations
 
 import argparse
 import tempfile
-from typing import TYPE_CHECKING, Any
+from typing import Any
 from unittest.mock import MagicMock, patch
+
+import pytest
 
 from causal_optimizer.cli import (
     _adapter_engine_kwargs,
@@ -21,9 +23,6 @@ from causal_optimizer.cli import (
     _cmd_resume,
     _cmd_run,
 )
-
-if TYPE_CHECKING:
-    import pytest
 from causal_optimizer.domain_adapters.base import DomainAdapter
 from causal_optimizer.types import (
     Constraint,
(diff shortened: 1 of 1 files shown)


datablogin:
# Claude Code Review

## Review Feedback

Now I have a complete picture. Here's my review:

---

## PR Review: CLI Feature Parity — Adapter Config Hooks & Override Flags

### Overall Assessment

This is a well-structured PR that closes a real gap: the CLI previously hardcoded `"objective"` and couldn't configure key engine parameters. The layered override design (engine default < adapter hook < CLI flag) is clean and correct. A few issues worth addressing:

---

### Bugs / Issues

**1. `best_result()` not using the configured objective/minimize (medium severity)**

In `cli.py:126` and `cli.py:152`, `engine.log.best_result()` is called with no arguments, which defaults to `objective_name="objective"` and `minimize=True`. But the user may have configured a different objective name or maximize mode. This means the "best" result printed after `run`/`resume` may be wrong — it could pick the minimum of the wrong metric.

```python
# cli.py:126 — should be:
best = engine.log.best_result(
    objective_name=objective_name,
    minimize=engine_kwargs.get("minimize", True),
)
```

Same issue at line 152 for `_cmd_resume`.

**Similarly in `_cmd_report` (line 171):** `log.best_result()` is called without passing `objective_name` or `minimize`. The report command doesn't have access to `minimize` at all — consider adding `--minimize`/`--maximize` to the report parser too, or at minimum pass `objective_name`.

**2. Report JSON format doesn't use `objective_name` (minor)**

In `_cmd_report`, the JSON output path (line 174-183) doesn't benefit from the `objective_name` flag at all. The table format shows `Best {objective_name}: ...` but the JSON format just dumps the raw `best_result`. This is an inconsistency — not a bug per se, but users comparing table vs JSON output will see different information.

---

### Code Quality

**3. `_apply_cli_overrides` uses `getattr` defensively — but why?**

Lines 86-94 use `getattr(args, "minimize", False)` etc. Since argparse populates the namespace from parser definitions, and `_apply_cli_overrides` is only called from `_cmd_run` and `_cmd_resume` which both use `_add_optimization_flags`, these attributes are guaranteed to exist. The `getattr` with defaults is defensive coding against a scenario that can't happen in normal use. It's not wrong, but it obscures the intent — a reader might wonder "can these attributes be missing?"

A comment like `# getattr for safety if called from a subparser without optimization flags` would clarify. Or just use `args.minimize` directly since the contract is well-defined.

**4. Duplicate boilerplate in tests (style nit)**

The test file has ~7 test classes that each set up an identical mock scaffolding (mock store with `__enter__`/`__exit__`, mock engine, tempfile). A shared fixture or helper would reduce the ~200 lines of boilerplate to maybe 50. Not blocking, but this is 516 lines of test for what amounts to ~80 lines of production code.

**5. Missing direct test for `_apply_cli_overrides`**

The function `_apply_cli_overrides` is tested only indirectly through `_cmd_run`/`_cmd_resume` integration tests. A direct unit test would be much simpler (no mocks needed) and would more clearly verify edge cases like:
- Neither `--minimize` nor `--maximize` set → no override
- All flags `None` → kwargs unchanged

---

### Backward Compatibility

**6. Existing adapters (`MarketingAdapter`, `MLTrainingAdapter`) don't override new hooks**

This is fine — the defaults (`objective_name="objective"`, `minimize=True`, `strategy="bayesian"`) match the previous hardcoded behavior. But it's worth noting that `MarketingAdapter.run_experiment` returns metrics like `{"revenue": ..., "conversion_rate": ...}` — there's no `"objective"` key. So with the defaults, `best_result()` will look for `"objective"` in those metrics and never find it. This was a pre-existing issue, not introduced by this PR, but the new hooks provide the mechanism to fix it (e.g., `MarketingAdapter.get_objective_name()` returning `"revenue"`).

**7. `_adapter_engine_kwargs` now always sets `objective_name`, `minimize`, `strategy`**

Before this PR, these weren't set in `engine_kwargs`, so `ExperimentEngine.__init__` would use its own defaults. After this PR, they're always explicitly set (from adapter hooks). The values are identical to the engine defaults, so no behavior change — but if the engine defaults ever diverge from adapter defaults, this explicit-setting behavior is actually better.

---

### Type Safety

**8. `get_strategy()` and `get_discovery_method()` return `str` — no validation**

The adapter methods return plain `str` with no validation. A subclass could return `get_strategy() -> "invalid_strategy"` and it would pass through silently until `ExperimentEngine.__init__` (presumably) rejects it. Consider using a `Literal` type:

```python
def get_strategy(self) -> Literal["bayesian", "causal_gp"]:
```

This would let mypy catch invalid returns in adapter subclasses. Same for `get_discovery_method()` → `Literal["correlation", "pc", "notears"] | None`.

---

### Test Coverage Gaps

- No test for the case where neither `--minimize` nor `--maximize` is passed (should preserve adapter's value)
- No test for report with `--format json` and `--objective-name`
- No test that `best_result()` is called with the correct objective name (currently it isn't — see bug #1)
- No test for the `_cmd_resume` path printing the correct objective name

---

### Summary

| Category | Verdict |
|----------|---------|
| Design | Clean layered override pattern (engine < adapter < CLI) |
| Bug #1 | `best_result()` ignores configured objective_name/minimize — should be fixed |
| Backward compat | Safe — defaults match previous behavior |
| Tests | Good coverage of happy paths, but heavy boilerplate and missing edge cases |
| Type safety | `str` return types for strategy/discovery could be `Literal` |

The main actionable item is bug #1 — `best_result()` should receive the configured `objective_name` and `minimize` values. Everything else is minor.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a comprehensive understanding. Here's my review:

---

## PR Review: CLI feature parity — adapter config hooks & override flags

### Overall Assessment

This is a well-structured PR that solves a real problem: the CLI was hardcoded to `"objective"` as the metric name and `minimize=True`, making it unusable for adapters with different metric names or maximization problems. The three-tier priority model (CLI flag > adapter hook > engine default) is clean and correct.

### Issues Found

#### 1. Bug: `_cmd_report` test missing `maximize`/`minimize` attributes (minor)

`tests/unit/test_cli_extended.py:459-464` — The `args` namespace for the report test doesn't include `minimize` or `maximize` attributes:

```python
args = argparse.Namespace(
    id="test-exp",
    db="test.db",
    format="table",
    objective_name="custom_obj",
    # missing: minimize=False, maximize=False
)
```

This works only because `_cmd_report` uses `getattr(args, "maximize", False)` with a safe default. It's not a functional bug, but it's inconsistent with every other test in the file that explicitly sets all attributes, and it makes the test fragile — if the code ever switches to `args.maximize` it would break silently. Should add `minimize=False, maximize=False` for consistency.

#### 2. Bug: `_cmd_report` ignores `--minimize` flag (real bug)

`causal_optimizer/cli.py:170`:
```python
minimize = not getattr(args, "maximize", False)
```

This logic only checks `--maximize`. If the user passes `--minimize` explicitly, it's ignored — `minimize` would already be `True` from the `not False` fallback. While functionally the default happens to be correct, it means `--minimize` is a dead flag on `report`. The `_add_optimization_flags` function isn't used for report (report has its own separate minimize/maximize group), but the report subparser does register `--minimize` as a flag that does nothing.

Consider using the same pattern as `_apply_cli_overrides`:
```python
if getattr(args, "maximize", False):
    minimize = False
elif getattr(args, "minimize", False):
    minimize = True
else:
    minimize = True  # default
```

This is semantically identical for the current logic but correctly documents the intent and would matter if the default ever changed.

#### 3. Design: `get_strategy()` return type should be a Literal or enum

`causal_optimizer/domain_adapters/base.py:52-57` — `get_strategy()` returns `str` but only `"bayesian"` and `"causal_gp"` are valid. The CLI uses `choices=["bayesian", "causal_gp"]` for validation, but adapter subclasses can return any string with no validation. Consider using `Literal["bayesian", "causal_gp"]` as the return type, which mypy would then enforce at override sites.

Same applies to `get_discovery_method()` — could be `Literal["correlation", "pc", "notears"] | None`.

#### 4. Design: `_cmd_resume` passes `search_space` to `ExperimentEngine.resume()` via `**engine_kwargs`

`causal_optimizer/cli.py:138-146` — `_adapter_engine_kwargs` includes `search_space` in the kwargs dict, and then this entire dict is spread into `ExperimentEngine.resume()`. The `resume()` classmethod also receives `runner` and `store` as explicit arguments plus `**engine_kwargs`. I verified this is compatible with the engine's `resume()` which accepts `**engine_kwargs`, so this is fine — just noting the flow is a bit indirect since `search_space`, `objective_name`, `minimize`, etc. all travel through the same kwargs dict.

#### 5. Test verbosity: heavy boilerplate repetition

The test file is 618 lines, but the actual behavioral coverage is relatively thin. The mock setup pattern (create adapter, mock store context manager, mock engine) is repeated ~10 times identically. A `pytest` fixture or helper method would cut 200+ lines while making the tests easier to maintain.

For example, a fixture like:
```python
@pytest.fixture
def mock_run_env():
    """Set up mocked store + engine for _cmd_run tests."""
    ...
```

This isn't blocking, but it's worth noting for maintainability.

#### 6. Asymmetry: `best_result()` call site after `with` block

`causal_optimizer/cli.py:125-129` — In `_cmd_run`, the `best_result()` call happens *after* the `with ExperimentStore(...)` block exits. The `engine.log` object is read after the store is closed. This isn't new to this PR (it existed before), but the PR does modify this code path. It works because `engine.log` is an in-memory object that doesn't need the store, but it's worth being aware of.

#### 7. Missing test: `--minimize` flag on `report` subcommand

The report subparser registers both `--minimize` and `--maximize` flags, but the test only covers `--objective-name` for reports. There's no test that verifies report behavior when `--maximize` is passed (i.e., that the correct best result is selected). Given issue #2 above, a test for this would have caught the dead `--minimize` flag.

### What's Good

- **Three-tier priority** (CLI > adapter > engine default) is well-designed and clearly documented in the docstring.
- **`_apply_cli_overrides` as a separate function** is clean and testable — the direct unit tests in `TestApplyCliOverrides` are the best tests in the file.
- **Backward compatibility** is preserved perfectly. All new adapter methods have defaults matching the existing engine defaults. Existing concrete adapters (`MarketingAdapter`, `MLTrainingAdapter`) need no changes.
- **`_add_optimization_flags` DRYs up** the shared flags between `run` and `resume` subparsers correctly.
- **Fixing the hardcoded `"objective"`** in `best_result()` and print output is a genuine bug fix.
- **`getattr` with defaults** in `_apply_cli_overrides` is defensive and correct — it handles cases where attributes might not exist on the namespace.

### Summary

| Category | Assessment |
|----------|-----------|
| Correctness | Good. One dead flag (`--minimize` on report), otherwise solid |
| Backward compat | Excellent. No breaking changes |
| Test coverage | Adequate but verbose. Missing report flag tests |
| Type safety | Could be tighter with Literal types on strategy/discovery |
| Architecture | Clean three-tier override pattern |

The PR is in good shape. The main actionable item is fixing the `_cmd_report` minimize logic to not silently ignore `--minimize`, and adding a test for `--maximize` on the report command.

---

PR #88, opened 2026-03-28
feat: null-signal control benchmark (Sprint 18)

### LINKED ISSUE #85: Null-signal control benchmark using permuted targets
## Problem

We have no way to verify that the optimizer does not manufacture false
wins from noise. A null-signal benchmark is needed as a negative control.

## What to do

Build a predictive null control that permutes the target column within
the train/validation region, destroying any real signal while preserving
marginal distributions. Run all strategies and verify no stable wins.

## Acceptance

1. No strategy shows consistent held-out improvement beyond noise
2. Any apparent wins are unstable across seeds
3. The suite report calls the result inconclusive or null
4. One regression test or smoke path exists

datablogin:
## Summary

- Add null-signal control benchmark that permutes the target column to destroy all signal, then runs all three strategies (random, surrogate_only, causal) to verify no false wins
- Benchmark module (`causal_optimizer/benchmarks/null_predictive_energy.py`) with `permute_target`, `run_null_strategy`, and `check_null_signal` functions
- Runner script (`scripts/null_energy_benchmark.py`) for CLI execution
- Smoke test and unit tests for the null benchmark (805 tests pass)
- Benchmark report documenting results: **PASS** -- no strategy wins on null data

## Benchmark Results

| Strategy | Mean Test MAE | Std |
|----------|--------------|-----|
| random | 3261.03 | 1.50 |
| surrogate_only | 3256.04 | 0.42 |
| causal | 3257.24 | 2.76 |

Max difference: 0.15% (well within the 2% noise threshold). The system correctly resists false discovery.

## Test plan

- [x] All 805 tests pass (`uv run pytest`)
- [x] Lint clean (`uv run ruff check .`)
- [x] Format clean (`uv run ruff format --check .`)
- [x] Type check clean (`uv run mypy causal_optimizer/`)
- [x] Smoke benchmark runs successfully
- [x] Full benchmark (budgets 20/40, seeds 0-2) completes with PASS verdict

Closes #85

<h3>Greptile Summary</h3>

This PR introduces a null-signal control benchmark that permutes the target column (destroying all temporal signal while preserving marginal distributions) and then runs all three strategies (random, `surrogate_only`, `causal`) to verify the optimizer does not manufacture false wins from noise. The new `null_predictive_energy.py` module slots cleanly alongside the existing `predictive_energy.py` benchmark, reusing `ValidationEnergyRunner`, `evaluate_on_test`, and `ExperimentEngine` without duplication.

All four issues raised in the prior review have been fully addressed:
- `_fmt_mean_std` now uses `ddof=1`, consistent with `check_null_signal`
- `NullSignalResult.verdict` now uses `\"ERROR\"` (not `\"WARN\"`) for the no-valid-results case, eliminating the previously ambiguous sentinel
- `main()` exits with code 1 on both `WARN` and empty-results paths, making it CI-gate safe
- `test_null_signal_check_high_variance_winner` now asserts `verdict.verdict == \"WARN\"` specifically, replacing the trivially-true `in {\"PASS\", \"WARN\"}` check

The benchmark logic itself looks sound: the per-budget check correctly catches false wins that would otherwise be masked by cross-budget averaging, and the `permute_target` function is applied to the full dataset before the train/val/test split, ensuring no temporal leakage.

Safe to merge — all four prior review findings are resolved and no new issues found.

All prior P1/P1-adjacent issues (inconsistent std formula, trivially-true test assertion, always-zero exit code, ambiguous WARN sentinel) are demonstrably fixed in this revision. The new code is well-tested (12 targeted unit tests covering all three verdict paths, per-budget masking, determinism, marginal preservation, and crash resilience) and follows established patterns from the existing benchmark suite. No remaining P0 or P1 findings.

| Filename | Overview |
|----------|----------|
| causal_optimizer/benchmarks/null_predictive_energy.py | New module implementing the null-signal benchmark: permute_target, run_null_strategy, check_null_signal, and NullSignalResult dataclass; clean implementation with proper ERROR/WARN/PASS distinction, per-budget and pooled checks, and ddof=1 std throughout. |
| scripts/null_energy_benchmark.py | CLI runner for the null benchmark; exits with code 1 on WARN or empty results, uses ddof=1 consistently in _fmt_mean_std, and passes the permuted DataFrame to split_time_frame before strategy runs. |
| tests/unit/test_null_benchmark.py | 12 tests covering permutation properties, check_null_signal verdict paths (PASS/WARN/ERROR), per-budget masking detection, monkeypatched crash path, and VALID_STRATEGIES export; high-variance WARN test now uses a specific assert rather than the previously trivially-true in-set check. |
| causal_optimizer/benchmarks/__init__.py | Exports NullSignalResult, check_null_signal, permute_target, and run_null_strategy from the new module; __all__ is updated in alphabetical order consistent with the rest of the file. |
| thoughts/shared/docs/sprint-18-null-signal-report.md | Sprint-18 benchmark report documenting PASS results with max 0.15% difference across strategies, well within the 2% threshold. |

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[Load energy DataFrame\nload_energy_frame] --> B[permute_target\nseed=99999]
    B --> C[split_time_frame\ntrain / val / test]
    C --> D{For each budget x seed x strategy}
    D --> E[run_null_strategy]
    E --> F{strategy == random?}
    F -- yes --> G[Random search loop\nsample_random_params]
    F -- no --> H[ExperimentEngine\nrun_loop]
    G --> I[evaluate_on_test]
    H --> I
    I --> J[PredictiveBenchmarkResult]
    J --> K[Collect results list]
    D -- next combo --> D
    K --> L[Write JSON output]
    K --> M[check_null_signal\nthreshold=2%]
    M --> N{strategy_means empty?}
    N -- yes --> O[verdict = ERROR\nexit 1]
    N -- no --> P[Per-budget check\nfor each budget]
    P --> Q[Pooled check\nacross all budgets]
    Q --> R{any winner?}
    R -- yes --> S[verdict = WARN\nexit 1]
    R -- no --> T[verdict = PASS\nexit 0]
```

<sub>Reviews (3): Last reviewed commit: ["fix: per-budget null verdict + review it..."](https://github.com/datablogin/causal-optimizer/commit/5d2c2db71df2fb3ecc3f9e08e47110900b51cefe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26665173)</sub>

### DIFF
--- a/causal_optimizer/benchmarks/null_predictive_energy.py
+++ b/causal_optimizer/benchmarks/null_predictive_energy.py
@@ -69,6 +69,12 @@ def permute_target(
     column, preserving all covariate structure and marginal distribution
     of the target.
 
+    .. note::
+
+        Call this **before** :func:`~causal_optimizer.benchmarks.predictive_energy.split_time_frame`
+        so that the permutation applies to the full dataset prior to the
+        train/val/test split.
+
     Args:
         df: Input DataFrame containing the target column.
         target_col: Name of the column to permute (default ``"target_load"``).
@@ -94,7 +100,7 @@ def permute_target(
 # ── Null benchmark strategy runner ────────────────────────────────────
 
 VALID_STRATEGIES: frozenset[str] = frozenset({"random", "surrogate_only", "causal"})
-_DEFAULT_CHECKPOINTS: list[int] = [5, 10, 20, 40, 80]
+_DEFAULT_CHECKPOINTS: tuple[int, ...] = (5, 10, 20, 40, 80)
 
 
 def run_null_strategy(
@@ -219,13 +225,19 @@ def check_null_signal(
     if no strategy is more than *threshold* (relative) better than the
     ``"random"`` baseline.  Returns WARN if any strategy appears to win.
 
+    The check is applied both **per-budget** and **pooled across budgets**.
+    If any single budget shows a false win exceeding *threshold*, the
+    overall verdict is WARN even if the pooled means cancel out.  This
+    prevents budget-level false wins from being masked by averaging.
+
     .. note::
 
         This is a simple threshold check on the mean-of-means, **not** a
         formal hypothesis test (e.g., paired t-test).  It is designed as
         a fast smoke check for the negative-control benchmark.  Strategies
         with zero valid results are flagged in the details list so callers
-        can detect incomplete runs.
+        can detect incomplete runs.  Results whose ``strategy`` is not in
+        *strategies* are silently ignored.
 
     Args:
         results: List of benchmark results (across strategies and seeds).
@@ -237,7 +249,7 @@ def check_null_signal(
     """
     details: list[str] = []
 
-    # Compute mean test MAE per strategy
+    # Compute mean test MAE per strategy (pooled across all budgets)
     strategy_maes: dict[str, list[float]] = {s: [] for s in strategies}
     for r in results:
         if r.strategy in strategy_maes and math.isfinite(r.test_mae):
@@ -283,19 +295,69 @@ def check_null_signal(
 
     best_improvement = (baseline_mae - best_mae) / baseline_mae if baseline_mae > 0 else 0.0
 
-    # Check if any strategy beats baseline by more than threshold
-    has_consistent_winner = False
+    # ── Per-budget check ──────────────────────────────────────────────
+    # Group results by budget and check each budget independently.
+    budgets_seen: set[int] = {r.budget for r in results if r.strategy in strategy_maes}
+    per_budget_winner = False
+
+    if len(budgets_seen) > 1:
+        details.append("--- per-budget breakdown ---")
+
+    for budget in sorted(budgets_seen):
+        budget_maes: dict[str, list[float]] = {s: [] for s in strategies}
+        for r in results:
+            if r.budget == budget and r.strategy in budget_maes and math.isfinite(r.test_mae):
+                budget_maes[r.strategy].append(r.test_mae)
+
+        budget_means: dict[str, float] = {}
+        for s, maes in budget_maes.items():
+            if maes:
+                budget_means[s] = float(np.mean(maes))
+
+        budget_baseline_key = (
+            "random"
+            if "random" in budget_means
+            else max(budget_means, key=lambda k: budget_means[k])
+            if budget_means
+            else None
+        )
+        if budget_baseline_key is None:
+            continue
+        budget_baseline = budget_means[budget_baseline_key]
+        if budget_baseline <= 0:
+            continue
+
+        for s, mean_mae in budget_means.items():
+            if s == budget_baseline_key:
+                continue
+            rel_improvement = (budget_baseline - mean_mae) / budget_baseline
+            if rel_improvement > threshold:
+                per_budget_winner = True
+                details.append(
+                    f"WARNING: budget={budget}: {s} beats {budget_baseline_key} "
+                    f"by {rel_improvement:.1%} (>{threshold:.0%} threshold)"
+                )
+            elif len(budgets_seen) > 1:
+                details.append(
+                    f"budget={budget}: {s} vs {budget_baseline_key} "
+                    f"delta={rel_improvement:.1%} (within threshold)"
+                )
+
+    # ── Pooled check ──────────────────────────────────────────────────
+    pooled_winner = False
     for s, mean_mae in strategy_means.items():
         if s == baseline_key:
             continue
         if baseline_mae > 0:
             rel_improvement = (baseline_mae - mean_mae) / baseline_mae
             if rel_improvement > threshold:
-                has_consistent_winner = True
+                pooled_winner = True
                 details.append(
-                    f"WARNING: {s} beats {baseline_key} by {rel_improvement:.1%} "
-                    f"(>{threshold:.0%} threshold) on null data"
+                    f"WARNING: pooled: {s} beats {baseline_key} by "
+                    f"{rel_improvement:.1%} (>{threshold:.0%} threshold) on null data"
                 )
+
+    has_consistent_winner = per_budget_winner or pooled_winner
 
     verdict: Literal["PASS", "WARN", "ERROR"]
     if has_consistent_winner:
--- a/tests/unit/test_null_benchmark.py
+++ b/tests/unit/test_null_benchmark.py
@@ -12,6 +12,7 @@ import pandas as pd
 import pytest
 
 from causal_optimizer.benchmarks.null_predictive_energy import (
+    VALID_STRATEGIES,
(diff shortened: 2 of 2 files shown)


datablogin:
@greptileai

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the PR. Here's my review:

---

## PR Review: feat: null-signal control benchmark (Sprint 18)

### Overall Assessment

This is a well-designed negative control benchmark. The core idea — permute targets to destroy signal, then verify no strategy "wins" — is sound methodology. The implementation is clean and follows the existing codebase patterns well. A few issues worth addressing:

---

### Issues

**1. `NullSignalResult` and `check_null_signal` not exported from `benchmarks/__init__.py`**

The new public types (`NullSignalResult`, `check_null_signal`, `permute_target`, `run_null_strategy`) are not added to `causal_optimizer/benchmarks/__init__.py` or its `__all__`. The existing benchmark module exports everything else (`PredictiveBenchmarkResult`, `CounterfactualBenchmarkResult`, etc.). This is an inconsistency — consumers importing from `causal_optimizer.benchmarks` won't discover the null-signal API.

**2. `_VALID_STRATEGIES` is duplicated across two files**

`null_predictive_energy.py:96` and `null_energy_benchmark.py:43` both define `_VALID_STRATEGIES: frozenset[str] = frozenset({"random", "surrogate_only", "causal"})`. If a strategy name ever changes or a new one is added, these will drift apart silently. Consider importing from one canonical location (the library module).

**3. `check_null_signal` uses mean-of-means without accounting for sample size or variance**

`null_predictive_energy.py:268` — The null-signal check compares mean test MAE per strategy to a fixed relative threshold (2%), but doesn't use a statistical test. With 3 seeds (as in the report) and the variance observed, even a proper t-test would have very low power. This is fine as a heuristic for now, but worth documenting that this is a threshold check, not a formal hypothesis test. The `strategies` parameter is also unused beyond filtering — if a strategy in the list has zero results, it's silently skipped rather than flagged.

**4. `verdict` field on `NullSignalResult` is a raw `str` rather than a `Literal` or enum**

`null_predictive_energy.py:51` — The `verdict` field accepts any string, but the module only produces `"PASS"`, `"WARN"`, or `"ERROR"`. Using `Literal["PASS", "WARN", "ERROR"]` would make the type contract explicit and catch typos at type-check time (mypy strict is enabled per CLAUDE.md).

**5. Smoke test only exercises the `"random"` strategy**

`test_null_benchmark.py:136` — `test_null_benchmark_smoke` only runs the `"random"` strategy path. The `"surrogate_only"` and `"causal"` paths through `run_null_strategy` (which construct an `ExperimentEngine`) are never exercised in tests. These paths have more complex setup (graph, descriptors, anytime metrics) and could fail silently. A parametrized smoke test over all three strategies would catch this.

**6. Missing test for `permute_target` with invalid column name**

The function raises `ValueError` on missing `target_col`, but there's no test for this error path. Minor, but easy to add for completeness.

**7. Missing test for `run_null_strategy` with invalid strategy name**

Same — `run_null_strategy` validates strategy and raises `ValueError`, but no test covers it.

---

### Minor / Stylistic

**8. Imports inside test function body**

`test_null_benchmark.py:142-148` — `run_null_strategy` and `split_time_frame` are imported inside the test function body rather than at module top. The other imports from the same modules are already at the top of the file. Move these to the top for consistency (unless this is intentional to defer the import for `@pytest.mark.slow`-skipped tests — if so, a comment would clarify intent).

**9. `_sanitize_for_json` in the script could use `np.ndarray` handling**

`scripts/null_energy_benchmark.py:47` — The sanitizer handles `np.integer`, `np.floating`, and `np.bool_` but not `np.ndarray`. If any nested value is an array (unlikely but possible from `dataclasses.asdict` on a result with array-containing fields), it would pass through un-sanitized and then `json.dump` would raise.

**10. Report: "n=6" in the mean table aggregates across budgets**

`sprint-18-null-signal-report.md:48-52` — The summary table reports n=6 per strategy, which is 2 budgets x 3 seeds. Aggregating across budgets is a design choice worth a brief note, since budget could plausibly affect the result (more optimization budget = more opportunity to overfit noise). The per-budget table above shows this isn't an issue here, but the aggregation should be explicit.

---

### What's Good

- **Correct permutation-before-split ordering** in both the script and the smoke test — this is the critical part of the null construction and it's done right.
- **`permute_target` copies the DataFrame** before mutating, avoiding side effects on the caller's data.
- **Deterministic seeds** throughout — the permutation seed (99999) is fixed and separate from the strategy seed, preventing accidental correlation.
- **Thorough marginal-preservation tests** — mean, std, min, max, covariate invariance are all checked.
- **The report analysis is substantive** — not just numbers, but interpretation of the validation-test gap pattern and surrogate convergence behavior.
- **`_sanitize_for_json`** correctly handles NaN/Inf → None conversion, which is necessary for `json.dump(allow_nan=False)`.
- **Clean separation** between the library module (testable logic) and the CLI script (I/O + orchestration).

---

datablogin:
I found one important acceptance gap in the null-signal check.

`check_null_signal()` currently pools all budgets together before deciding PASS/WARN. That means a strategy can show a clear false win at one budget and a clear loss at another, and the benchmark can still average out to `PASS`. I reproduced this directly with budget 20 where `causal` is 10% worse than random and budget 40 where `causal` is 10% better than random; the current function returns `PASS` because the pooled mean lands back at parity.

That is too forgiving for a negative control. The Sprint 18 plan says we should run across seeds and budgets and treat apparent wins as unstable or inconclusive. A budget-specific promotable win should not be able to hide inside pooled means.

Suggested fix:
1. evaluate the null verdict per budget (or per `(budget, strategy)` slice against random),
2. return `WARN` if any budget exceeds the false-win threshold, and
3. add a unit test for the “one budget wins, another loses, pooled mean passes” masking case.

datablogin:
I reran the pooled-budget masking repro on current HEAD (`9d37f1f`), and the original issue is still present.

I used a case where `causal` is 10% worse than random at budget 20 and 10% better than random at budget 40. The current `check_null_signal()` still returns:

```
NullSignalResult(verdict='PASS', has_consistent_winner=False, ...)
```

because it is still pooling all budgets together before deciding PASS/WARN. The new tests cover zero-result strategies, but they do not cover the masking case from the review.

So the negative control is still too forgiving: a budget-specific false win can still disappear inside pooled means.

Please keep the fix aligned to the original request:
1. evaluate the null verdict per budget (or per `(budget, strategy)` slice against random),
2. return `WARN` if any budget exceeds the false-win threshold, and
3. add a unit test for the “one budget wins, another loses, pooled mean passes” case.

Issue #153, opened 2026-04-10
Sprint 29: implement adaptive causal guidance under Ax-primary gates


datablogin:
## Goal

Implement one narrow optimizer-core intervention based on the Sprint 29 trajectory diagnosis.

The objective is adaptive causal guidance, not universal causal wins.

## Contract

Use the prompt:

`/Users/robertwelborn/Projects/causal-optimizer/thoughts/shared/prompts/sprint-29-adaptive-causal-guidance.md`

Start only after the Sprint 29 trajectory diagnosis findings are available.

## Required Output

`thoughts/shared/docs/sprint-29-adaptive-causal-guidance-report.md`

## Acceptance Criteria

1. Implements exactly one narrow intervention justified by the trajectory diagnosis.
2. Preserves alignment-only / soft-causal behavior where expected.
3. Does not reopen balanced Ax reranking.
4. Does not add new benchmark families.
5. Adds focused tests for trigger behavior, no-op behavior, and relevant robustness cases.
6. Documents the exact mechanism changed and the risks the regression gate must check.
7. Passes CI and review gates.

Issue #152, opened 2026-04-10
Sprint 29: diagnose interaction and dose-response optimizer trajectories under Ax


datablogin:
## Goal

Diagnose why causal guidance is not yet cleanly better on the two rows that define the next optimizer-core frontier:

1. interaction policy
2. dose-response clinical

This is a diagnosis-only issue. Do not change optimizer behavior here.

## Contract

Use the prompt:

`/Users/robertwelborn/Projects/causal-optimizer/thoughts/shared/prompts/sprint-29-trajectory-diagnosis.md`

Read first:

1. `README.md`
2. `thoughts/shared/plans/20-sprint-29-recommendation.md`
3. `thoughts/shared/plans/07-benchmark-state.md`
4. `thoughts/shared/docs/sprint-28-backend-baseline-scorecard.md`
5. `thoughts/shared/docs/sprint-28-ax-regression-gate-report.md`

## Required Output

`thoughts/shared/docs/sprint-29-trajectory-diagnosis-report.md`

## Acceptance Criteria

1. Compares causal vs surrogate_only on interaction and dose-response.
2. Separates Ax-primary evidence from RF-secondary evidence.
3. Identifies whether the issue appears to be over-pruning, missed interactions, causal pressure timing, insufficient statistical power, or surrogate behavior.
4. Recommends one narrow follow-up intervention, or explains why not to intervene yet.
5. Makes no optimizer-core code changes.
6. Passes CI and review gates.

PR #159, opened 2026-04-10
docs: Sprint 29 interaction ablation (EXPLORATION WEIGHTING IS PRIMARY CAUSE)

datablogin:
## Summary

- 5-arm ablation on interaction policy benchmark under Ax/BoTorch primary path
- **Verdict: exploration weighting (causal_exploration_weight=0.3) is the primary cause** of the B20 catastrophe
- graph_only (w=0.0, s=0.0) is the best arm at every budget: B80 regret 1.80 (std 0.02), beating surrogate_only (2.18)
- Removing exploration weighting alone eliminates the B20 catastrophe (13.83 → 4.76)
- Alignment bonus amplifies (+6.06 at B20) but does not independently cause failure (+0.63 when exploration weighting is off)
- Causal graph's variable pruning helps when early pressure is removed
- Recommends setting causal_exploration_weight=0.0 as default for #153
- Includes reproducible ablation script (scripts/interaction_ablation.py)

Addresses #152

## Test plan

- [ ] Lint passes (`ruff check .`)
- [ ] Format passes (`ruff format --check .`)
- [ ] All 1001 unit tests pass
- [ ] Ablation script runs and produces consistent results
- [ ] Report answers all required questions from the prompt contract

<h3>Greptile Summary</h3>

Adds a 5-arm ablation script and corresponding analysis report isolating the effect of `causal_exploration_weight` (0.0 vs 0.3) and `causal_softness` (0.0 vs 0.5) on the interaction policy benchmark across 3 budgets × 10 seeds, concluding that exploration weighting is the primary cause of the B20 regression and recommending `causal_exploration_weight=0.0` as the new default for #153.

The ablation logic is correct: `InteractionPolicyScenario.generate()` re-seeds from `_seed=0` on every call, ensuring identical train/test splits across all 150 runs; the previous comment about missing arm-name validation is addressed at lines 145–147; and the `surrogate_only` arm correctly ignores the graph-related hyperparameters when `causal_graph=None` (confirmed in `suggest.py`). Report statistics were cross-verified against the per-seed table and are internally consistent.

Safe to merge — docs-and-script PR with no code-correctness issues found.

All findings are at P2 or lower. The script logic is correct, the previously raised validation concern is addressed, and report statistics are internally consistent. No blocking issues remain.

| Filename | Overview |
|----------|----------|
| scripts/interaction_ablation.py | New ablation script: 5-arm × 3-budget × 10-seed harness. Logic is correct — generate() re-seeds from _seed=0 on every call so data is deterministic across all 150 runs; surrogate_only arm correctly ignores causal_exploration_weight (verified: suggest.py line 245 short-circuits on causal_graph is None); previous validation concern is addressed at lines 145–147. |
| thoughts/shared/docs/sprint-29-interaction-ablation-report.md | Analysis report documenting the 5-arm ablation results. All reported means, standard deviations, and per-arm differences were cross-checked against the per-seed table and are internally consistent. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: align executive summary item 4 with..."](https://github.com/datablogin/causal-optimizer/commit/0df0700e53d9a1b62da88238b264531ab257400c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28040490)</sub>

### DIFF
--- a/scripts/interaction_ablation.py
+++ b/scripts/interaction_ablation.py
@@ -142,6 +142,9 @@ def main() -> None:
     budgets = [int(b) for b in args.budgets.split(",")]
     seeds = [int(s) for s in args.seeds.split(",")]
     arm_names = [a.strip() for a in args.arms.split(",")]
+    for name in arm_names:
+        if name not in ARMS:
+            parser.error(f"Unknown arm: {name}. Options: {','.join(ARMS.keys())}")
 
     df = pd.read_parquet(args.data_path)
     scenario = InteractionPolicyScenario(covariates=df, seed=0)
--- a/scripts/interaction_ablation.py
+++ b/scripts/interaction_ablation.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+"""Interaction policy ablation: early causal pressure components.
+
+Runs the interaction policy benchmark with different combinations of
+causal_exploration_weight and causal_softness to isolate which
+component of early causal pressure causes the B20 penalty.
+
+Arms:
+  1. surrogate_only (no causal graph)
+  2. causal_default (weight=0.3, softness=0.5)
+  3. no_exploration (weight=0.0, softness=0.5)
+  4. no_alignment (weight=0.3, softness=0.0)
+  5. graph_only (weight=0.0, softness=0.0)
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import time
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+
+from causal_optimizer.benchmarks.interaction_policy import (
+    InteractionPolicyRunner,
+    InteractionPolicyScenario,
+    evaluate_interaction_policy,
+)
+from causal_optimizer.benchmarks.provenance import collect_provenance
+from causal_optimizer.engine import ExperimentEngine
+
+logging.basicConfig(level=logging.WARNING)
+
+ARMS: dict[str, dict] = {
+    "surrogate_only": {
+        "use_graph": False,
+        "causal_exploration_weight": 0.3,  # irrelevant without graph
+        "causal_softness": 0.5,  # irrelevant without graph
+    },
+    "causal_default": {
+        "use_graph": True,
+        "causal_exploration_weight": 0.3,
+        "causal_softness": 0.5,
+    },
+    "no_exploration": {
+        "use_graph": True,
+        "causal_exploration_weight": 0.0,
+        "causal_softness": 0.5,
+    },
+    "no_alignment": {
+        "use_graph": True,
+        "causal_exploration_weight": 0.3,
+        "causal_softness": 0.0,
+    },
+    "graph_only": {
+        "use_graph": True,
+        "causal_exploration_weight": 0.0,
+        "causal_softness": 0.0,
+    },
+}
+
+
+def run_one(
+    scenario: InteractionPolicyScenario,
+    arm_name: str,
+    arm_cfg: dict,
+    budget: int,
+    seed: int,
+) -> dict:
+    """Run a single arm/budget/seed combination."""
+    t_start = time.monotonic()
+
+    data = scenario.generate()
+    n = len(data)
+    opt_end = int(n * 0.8)
+    val_data = data.iloc[:opt_end].reset_index(drop=True)
+    test_data = data.iloc[opt_end:].reset_index(drop=True)
+
+    space = scenario.search_space()
+    runner = InteractionPolicyRunner(val_data, scenario.treatment_cost)
+
+    graph = scenario.causal_graph() if arm_cfg["use_graph"] else None
+    engine = ExperimentEngine(
+        search_space=space,
+        runner=runner,
+        causal_graph=graph,
+        objective_name="objective",
+        minimize=True,
+        seed=seed,
+        max_skips=0,
+        causal_exploration_weight=arm_cfg["causal_exploration_weight"],
+        causal_softness=arm_cfg["causal_softness"],
+    )
+    engine.run_loop(budget)
+    best_result = engine.log.best_result("objective", minimize=True)
+    best_params = best_result.parameters if best_result is not None else None
+
+    if best_params is not None:
+        policy_value, decision_error = evaluate_interaction_policy(
+            test_data, best_params, scenario.treatment_cost
+        )
+    else:
+        policy_value = 0.0
+        oracle_treat = test_data["true_treatment_effect"].values > scenario.treatment_cost
+        decision_error = float(np.mean(oracle_treat))
+
+    oracle_value = scenario.oracle_policy_value(test_data)
+    regret = oracle_value - policy_value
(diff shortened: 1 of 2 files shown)


datablogin:
# Claude Code Review

## Review Feedback

Good — the script is API-compliant. Here's my review:

---

## PR Review: Sprint 29 Interaction Ablation

### Overall Assessment

This is a clean docs + tooling PR: one experiment script and one report. The script follows established patterns (matches `dose_response_benchmark.py`, `b80_seed_diagnostics.py`), uses correct APIs, and the report is well-structured with proper evidence/speculation separation. The ablation design is methodologically sound (2x2 factorial + baseline = 5 arms).

### Script: `scripts/interaction_ablation.py`

**Minor issues:**

1. **Bare `dict` in type annotations** (lines 37, 69, 72, 149): The module-level `ARMS` is typed `dict[str, dict]` and the function signatures use bare `dict`. Other scripts in the repo use `dict[str, Any]` for these cases. Consider `dict[str, Any]` for the inner dicts to be explicit — though this is cosmetic since `scripts/` likely isn't under mypy strict.

2. **`n` is computed but only used to derive `opt_end`** (line 77-78): Minor readability nit — `opt_end = int(len(data) * 0.8)` would be clearer than assigning `n` for a single use.

3. **Scenario seed is hardcoded to 0** (line 147) while the per-run seed varies. This is correct for reproducibility (same data splits across arms), but worth a comment since it could look like a bug to a future reader. The existing benchmark scripts have the same pattern though, so this is consistent.

4. **No `--arms` validation** (line 144): If someone passes `--arms bad_name`, they get a `KeyError` at line 156 with no helpful message. A quick check like:
   ```python
   for name in arm_names:
       if name not in ARMS:
           parser.error(f"Unknown arm: {name}. Options: {','.join(ARMS.keys())}")
   ```
   would improve UX. Not blocking.

5. **Line 139 is 97 chars** — fine under the 100-char limit, just noting it's close.

**No bugs found.** The data flow (generate → split → engine → evaluate → collect) is correct. The `best_params is None` fallback at line 105-108 is a reasonable defensive path.

### Report: `sprint-29-interaction-ablation-report.md`

**Strengths:**
- Excellent ablation design — the 2x2 factorial isolates each component cleanly
- Evidence vs Speculation section (Section 6) is exactly right
- Per-seed detail (Section 3) makes the catastrophe pattern visceral
- MWU tests properly labeled as two-sided (consistent with project feedback)
- Success criteria for #153 are concrete and testable

**One thing to verify:**
- Section 4b: "Difference (s=0.5 - s=0.0)" labels are confusing because the rows are ordered `no_alignment (s=0.0)` then `causal_default (s=0.5)`, but the difference is computed as `causal_default - no_alignment`. The label says `s=0.5 - s=0.0` which is correct mathematically, but the visual ordering (s=0.0 row first, s=0.5 row second) makes you expect the subtraction to go the other direction. Consider swapping row order or adding a note. Same pattern appears in the second comparison table in 4b.

### Not in scope but worth noting

The report recommends changing defaults (`causal_exploration_weight=0.0`, `causal_softness=0.0`) in issue #153. That's a code change that will need its own PR with regression gates — this PR correctly scopes itself to documenting the evidence.

### Verdict

**Approve.** Clean docs PR with a well-designed experiment. The minor script nits (bare `dict` types, no `--arms` validation) are non-blocking for a `scripts/` utility.

---

datablogin:
@greptileai

datablogin:
Fact-check review on current HEAD (`3e95218`). I recomputed the ablation tables from the local artifact:

`/Users/robertwelborn/Projects/_local/causal-optimizer/artifacts/sprint-29-interaction-ablation/ablation_results.json`

The quantitative claims check out from my side:
1. provenance confirms `optimizer_path: ax_botorch`
2. B20 means/stds match the report (`surrogate_only 5.44/1.83`, `causal_default 13.83/5.06`, `no_exploration 4.76/1.77`, `no_alignment 7.77/2.02`, `graph_only 4.13/2.07` using population std)
3. B40/B80 tables and the MWU p-values match (`graph_only` B40 `p=0.018`, `causal_default` B80 `p=0.014`, etc.)
4. the isolated deltas also match (`-9.07`, `+6.06`, `+0.63`)

I do have one substantive report-side concern before I’d sign off:

1. **The `graph_only` mechanism is overstated / cites the wrong path.** Sections 4c and 7 say the graph-only improvement shows that “variable pruning” / “candidate generation targeting” is helping, and Section 4c explicitly points to `_suggest_surrogate()`. But this study is Ax-primary (`optimizer_path: ax_botorch`), and on the Ax path with `causal_graph is not None` and `causal_softness=0.0`, `_suggest_bayesian()` still runs in soft mode and sets `ax_focus=None`, so Ax is optimizing all variables during the optimization phase; `_suggest_surrogate()` is only the RF fallback path and is not what produced these results here. In other words: the ablation shows that `graph_only` beats `surrogate_only`, but it does **not** establish that the benefit came from graph-based variable pruning during Ax optimization. Please soften that mechanism claim and the downstream recommendation text to something like “graph_only is empirically better on this row, but the exact Ax-path mechanism is not isolated by this ablation.”

A smaller script issue also remains, same one Greptile flagged:

2. `scripts/interaction_ablation.py` should validate `--arms` before `ARMS[arm_name]` so invalid input fails with `parser.error(...)` instead of a raw `KeyError`.

Once the `graph_only` mechanism wording is tightened, the ablation evidence itself looks solid from my side.

datablogin:
@greptileai

datablogin:
Re-reviewed current HEAD (`aec7ce1`). The `--arms` validation guard is fixed, and the body of the report now correctly softens the `graph_only` mechanism claim.

One small report-side consistency issue still remains before I’d sign off:

1. In the executive summary / key findings, item 4 still says:
   > "the causal graph's variable pruning helps when it does not distort exploration or optimization"

   That is stronger than the updated Section 4c / Section 7 wording. The body now correctly says the exact Ax-path mechanism is not isolated by this ablation, and that what we can claim is empirical: `graph_only` beats `surrogate_only` when early pressure is removed. I would make the key-finding bullet match the body, e.g. “Graph-only causal guidance (no early pressure) outperforms surrogate-only; the exact Ax-path mechanism is not isolated by this ablation.”

The ablation math still checks out from my side. After that one stale summary line is aligned with the body, I’m good to sign off.

datablogin:
Re-reviewed current HEAD (`0df0700`). The remaining summary-line inconsistency is fixed from my side.

What I checked:
1. The executive summary / key findings now matches the body: `graph_only` is described as empirically better, with the exact Ax-path mechanism explicitly left unisolated by this ablation.
2. The `--arms` validation guard is present, so invalid arm names now fail with `parser.error(...)` instead of a raw `KeyError`.
3. The quantitative ablation claims from my earlier fact-check remain aligned with the local artifact.
4. The visible CI checks on this push are green.

No blocking findings remain from my review. This is ready from my side.

datablogin:
@greptileai

PR #167, opened 2026-04-14
docs: Sprint 31 Hillstrom benchmark contract

datablogin:
## Summary

- Locks Hillstrom as the first non-energy real-data benchmark, with `Womens E-Mail vs No E-Mail` as the primary binary slice and pooled `Any E-Mail vs No E-Mail` as a secondary reference slice.
- Maps Hillstrom columns into `MarketingLogAdapter` with no adapter code changes — one small `HillstromLoader` wrapper handles the cost/segment/propensity synthesis and collapses the degenerate channel-share variables.
- Defines the 10-seed / B20-B40-B80 / null-control evidence contract and queues Criteo Uplift (scale-up) and Open Bandit (multi-action architecture) as follow-ons.
- Patches two precision issues in the Sprint 30 ERCOT reality report: COAST causal-vs-random narrative and the 10-seed rerun cost.

## Why Hillstrom first

Smallest credible non-energy answer. Randomized RCT (known propensities), binary-collapsible onto the existing adapter, small enough to iterate fast, and creates a clean ramp into Criteo and Open Bandit. A negative result is still diagnosable because the dataset is open and well studied.

## Scope discipline

- No changes to \`causal_optimizer/domain_adapters/marketing_logs.py\`
- No new adapter class
- No multi-action work (deferred to Sprint 33+ after Criteo)
- No Sprint 29 default tuning on the first non-energy run

## Test plan

- [ ] Human review of the launch recommendation (Womens-slice primary vs pooled primary)
- [ ] Human review of the column-mapping decisions (fixed-cost assumption, constant channel, `history_segment` → `segment` bucket map)
- [ ] Human review of the search-space degeneracy handling (`email_share=1.0`, `social_share_of_remainder=0.0`)
- [ ] Human review of the null-control definition (permuted outcome, 2% tolerance)
- [ ] Human review of the Sprint 30 ERCOT report precision patches
- [ ] Run gauntlet on the PR (greploop + claudeloop + check-pr)

This PR is docs-only. No code gates (lint/format/tests/coverage/mypy) run against markdown. Do not merge without explicit human approval per the project merge policy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This docs-only PR adds the Sprint 31 Hillstrom benchmark contract — a comprehensive planning document for the first non-energy real-data benchmark — and patches two precision issues in the Sprint 30 ERCOT reality report (COAST narrative B40/B80 ordering and the 10-seed rerun count).

[...]

PR #169, opened 2026-04-15
feat: Sprint 31 Hillstrom benchmark harness

### LINKED ISSUE #168: Sprint 31: implement the Hillstrom benchmark harness
## Why

Sprint 31 now has a locked Hillstrom contract, but the repo does not yet have the implementation surface to execute it. We need the first non-energy real-data benchmark harness to exist in code, under the exact constraints established in the merged contract, so we can test generalization beyond ERCOT without reopening benchmark-design questions mid-implementation.

## Primary Contract

- `thoughts/shared/docs/sprint-31-hillstrom-benchmark-contract.md`
- `thoughts/shared/plans/22-sprint-31-generalization-research-plan.md`
- `thoughts/shared/docs/sprint-30-general-causal-portability-brief.md`

## Scope

Implement the Hillstrom benchmark harness and smoke-test surface under the locked contract.

Required implementation shape:

1. add a narrow `HillstromLoader` wrapper that reuses `MarketingLogAdapter` unchanged
2. support the primary slice: `Womens E-Mail` vs `No E-Mail`
3. support the secondary pooled slice: `Any E-Mail` vs `No E-Mail`
4. project the adapter's 14-edge prior graph down to the active Hillstrom nodes defined in the contract
5. add the permuted-outcome null-control path for the primary slice
6. provide a runnable benchmark entrypoint / scenario surface for Hillstrom
7. add a small cached fixture slice for CI and deterministic smoke coverage

## Deliverables

1. Hillstrom loader / reshaping module
2. Hillstrom benchmark scenario and/or script entrypoint
3. small committed fixture slice for CI, with the full dataset remaining local-only
4. focused tests for:
   - primary-slice propensity = `0.5`
   - pooled-slice propensity = `2.0 / 3.0`
   - projected prior graph matches the contract
   - null-control setup runs on the fixture
5. provenance fields covering source dataset hash or version, slice type, and projected-graph usage
6. any minimal doc updates needed so the implementation path is discoverable

## Acceptance Criteria

1. `MarketingLogAdapter` is reused without modifying `causal_optimizer/domain_adapters/marketing_logs.py`
2. the implementation pre-bakes the frozen Hillstrom dimensions exactly as the contract requires
3. the benchmark can execute a 5-seed smoke run on the fixture without reopening contract decisions
4. the two propensity invariants are covered by tests and cannot silently regress
5. the projected prior graph is explicit and recorded in provenance
6. no multi-action / three-arm policy support is introduced in this issue
7. no optimizer-core tuning is introduced in this issue

## Out Of Scope

1. full three-arm multi-action support
2. Criteo or Open Bandit implementation
3. Sprint 29 default tuning
4. publishing a final Hillstrom benchmark verdict from smoke-test evidence alone

## Recommended First Pass

1. implement loader and graph projection
2. stand up fixture-backed smoke coverage
3. add a runnable benchmark script or scenario class
4. verify the primary, pooled, and null-control paths locally
5. only after the harness is stable, use it for the full benchmark run/report in follow-on work

datablogin:
Closes #168.

## Summary

Implements the Sprint 31 Hillstrom benchmark under the launch contract locked in `thoughts/shared/docs/sprint-31-hillstrom-benchmark-contract.md` (PR #167, still pending review). Reuses `MarketingLogAdapter` unchanged via a narrow loader + wrapped runner.

- `causal_optimizer/benchmarks/hillstrom.py` — `HillstromLoader` (`load_hillstrom_slice`), `HillstromSliceType` enum, active 3-variable search space, 7-edge projected prior graph, `HillstromPolicyRunner` that pre-bakes the three frozen dimensions, permuted-outcome null-control helpers, `HillstromScenario` / `HillstromBenchmarkResult`.
- `scripts/hillstrom_benchmark.py` — CLI runner over (slice, budget, seed, strategy) with optional `--null-control` pass and Hillstrom-specific provenance stamping (projected-graph edge list, per-slice null baselines, full dataset hash).
- `tests/fixtures/hillstrom_fixture.csv` — 300-row Hillstrom-shaped synthetic fixture (three balanced arms) with zero-inflated spend and realistic arm-dependent response rates.
- Tests: 34 unit + 11 integration, 100% coverage on the new module.

## Contract fidelity

| Contract invariant | How the code enforces it |
|-------------------|--------------------------|
| Primary propensity = exactly `0.5` | `HILLSTROM_PRIMARY_PROPENSITY = 0.5`; loader fills the column with `np.full(..., 0.5)`; unit test asserts bitwise equality on every row |
| Pooled propensity = exactly `2/3` | `HILLSTROM_POOLED_PROPENSITY = 2.0 / 3.0`; loader fills the column with `np.full(..., 2.0/3.0)`; unit test asserts bitwise equality and `!= 0.5` |
| Frozen params pre-baked | `HILLSTROM_FROZEN_PARAMS = {email_share: 1.0, social_share_of_remainder: 0.0, min_propensity_clip: 0.01}`; `HillstromPolicyRunner._forward_params` injects all three into every forwarded call without mutating the caller's dict |
| 7-edge projected prior graph | `hillstrom_projected_prior_graph()` returns the 7 explicit edges; unit tests assert: exact edge set, no frozen variable as tail/node, strict-subset of the full 14-edge adapter graph |
| Null baseline = `μ = mean(spend)` on the unshuffled frame | `hillstrom_null_baseline(real_slice)` returns a scalar once at scenario construction time; unit test confirms the baseline is invariant under any permutation seed |
| `history_segment` bucket map uses raw CSV strings with numeric prefix | `_HISTORY_SEGMENT_TO_SEGMENT` keyed on `"1) $0 - $100"`, etc. |
| No changes to `domain_adapters/marketing_logs.py` | Loader wraps an unmodified `MarketingLogAdapter` instance; `git diff main -- causal_optimizer/domain_adapters/marketing_logs.py` is empty |

## Smoke verification

- `uv run pytest tests/unit/test_hillstrom_benchmark.py tests/integration/test_hillstrom_benchmark.py` → **45/45 pass** in 0.69s.
- `uv run pytest -m "not slow"` → **1096 passed**, 10 skipped, 0 failures.
- `--cov=causal_optimizer` → **100% line coverage** on `causal_optimizer/benchmarks/hillstrom.py` (155/155 statements).
- `uv run mypy causal_optimizer/` → clean.
- `uv run ruff check . && uv run ruff format --check .` → clean.
- `scripts/hillstrom_benchmark.py --data-path tests/fixtures/hillstrom_fixture.csv --slices primary --budgets 20 --seeds 0,1,2,3,4 --strategies random,surrogate_only,causal --null-control` completes in **~1m48s** under `ax_botorch` (30 adapter evaluations). Observed results on the fixture:
  - Real primary slice: causal / surrogate_only `4.24 ± 0.35` vs random `3.01 ± 0.22`.
  - Null control: all strategies `~3.2–3.3` vs `μ=2.45`. This is noisy on the 300-row fixture with only 18 nonzero spend rows — expected behavior, and exactly the scenario the contract's fallback ladder in Section 5g is designed for. **Not a benchmark verdict.** The contract explicitly cautions against drawing conclusions from fixture-sized smoke runs; a real benchmark run against the full Hillstrom CSV is a separate sprint-task.

## What this PR does not do

- Full Hillstrom benchmark runs (those need the full CSV, local-only).
- Multi-action three-arm policy — deferred to a later sprint per contract Section 4c.
- Any adapter code changes.
- Any optimizer-core default tuning.

## Test plan

- [ ] Unit tests: `uv run pytest tests/unit/test_hillstrom_benchmark.py -v`
- [ ] Integration tests: `uv run pytest tests/integration/test_hillstrom_benchmark.py -v`
- [ ] Full fast suite regression: `uv run pytest -m "not slow"`
- [ ] Coverage: `uv run pytest tests/unit/test_hillstrom_benchmark.py tests/integration/test_hillstrom_benchmark.py --cov=causal_optimizer --cov-report=term-missing` — expect 100% on `causal_optimizer/benchmarks/hillstrom.py`
- [ ] CLI smoke: `uv run python scripts/hillstrom_benchmark.py --data-path tests/fixtures/hillstrom_fixture.csv --slices primary --budgets 20 --seeds 0 --strategies random --output /tmp/hillstrom_smoke.json`
- [ ] Human approval before merge, per CLAUDE.md hard merge rule.

<h3>Greptile Summary</h3>

This PR implements the Sprint 31 Hillstrom benchmark harness as specified by the locked contract doc. It adds a loader (`load_hillstrom_slice`), a frozen-param injecting runner (`HillstromPolicyRunner`), three strategies (random / surrogate_only / causal), and a null-control permutation path — all wired together in `HillstromScenario` — with a 300-row synthetic fixture and 45 tests (34 unit + 11 integration) at 100% coverage. The three issues flagged in the prior review round (duplicated `_VALID_STRATEGIES`, misleading `--null-control` help text, and the `params`-unused secondary-outcomes signature) are all resolved in this version.

Safe to merge — all contract invariants are correctly implemented and the three prior review comments are resolved.

All remaining findings are P2 style suggestions (split import block, mutable module-level dict). No correctness, data-integrity, or behavioral issues were found. Contract fidelity is solid: propensity pins, frozen-param injection, 7-edge projected graph, and null-control baseline are all implemented and pin-tested.

| Filename | Overview |
|----------|----------|
| causal_optimizer/benchmarks/hillstrom.py | Core benchmark module: contract constants, loader, projected prior graph, policy runner, null-control helpers, and scenario — well-structured with good defensive guards. Minor: module-level HILLSTROM_FROZEN_PARAMS dict is mutable. |
| scripts/hillstrom_benchmark.py | CLI runner over the strategy/budget/seed grid with JSON provenance output; imports VALID_STRATEGIES from the benchmark module (prior duplication issue resolved); null-control help text now accurately reflects the b <= 40 fallback logic. |
| causal_optimizer/benchmarks/__init__.py | Re-exports all new Hillstrom public symbols; uses two separate import blocks from the same hillstrom module — harmless but could be merged. |
| tests/unit/test_hillstrom_benchmark.py | 34 unit tests locking all contract invariants (slice mapping, propensity pins, edge set, search space bounds, frozen-param injection, null permutation); includes cache-clearing discipline around the monkeypatched lru_cache test. |
| tests/integration/test_hillstrom_benchmark.py | 11 integration tests exercising the full engine loop path for all three strategies on both slices and the null-control path; deliberately small budgets (8 experiments) for CI speed. |
| tests/unit/test_hillstrom_benchmark_script.py | Covers CLI helpers and an end-to-end main() smoke test; includes a regression guard confirming the script's VALID_STRATEGIES is the same object as the module's. |
| tests/fixtures/hillstrom_fixture.csv | 300-row synthetic fixture with three balanced arms, zero-inflated spend, and arm-dependent response rates. |

</details>

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%202%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%202%0Acausal_optimizer%2Fbenchmarks%2F__init__.py%3A23-25%0A**Two%20import%20blocks%20from%20the%20same%20module**%0A%0A%60VALID_STRATEGIES%20as%20HILLSTROM_VALID_STRATEGIES%60%20is%20split%20into%20its%20own%20block%20solely%20because%20it's%20a%20renamed%20import.%20Merging%20it%20into%20the%20first%20block%20is%20cleaner%20and%20avoids%20two%20separate%20module%20loads%20%28Python%20deduplicates%20them%2C%20but%20the%20visual%20split%20suggests%20a%20distinction%20that%20doesn't%20exist%20at%20runtime%29%3A%0A%0A%60%60%60suggestion%0Afrom%20causal_optimizer.benchmarks.hillstrom%20import%20%28%0A%20%20%20%20HILLSTROM_FROZEN_PARAMS%2C%0A%20%20%20%20HILLSTROM_POOLED_PROPENSITY%2C%0A%20%20%20%20HILLSTROM_PRIMARY_PROPENSITY%2C%0A%20%20%20%20VALID_STRATEGIES%20as%20HILLSTROM_VALID_STRATEGIES%2C%0A%20%20%20%20HillstromBenchmarkResult%2C%0A%20%20%20%20HillstromPolicyRunner%2C%0A%20%20%20%20HillstromScenario%2C%0A%20%20%20%20HillstromSliceType%2C%0A%20%20%20%20hillstrom_active_search_space%2C%0A%20%20%20%20hillstrom_null_baseline%2C%0A%20%20%20%20hillstrom_projected_prior_graph%2C%0A%20%20%20%20load_hillstrom_slice%2C%0A%20%20%20%20permute_hillstrom_spend%2C%0A%29%0A%60%60%60%0A%0A%23%23%23%20Issue%202%20of%202%0Acausal_optimizer%2Fbenchmarks%2Fhillstrom.py%3A168-172%0A**%60HILLSTROM_FROZEN_PARAMS%60%20is%20a%20mutable%20module-level%20dict**%0A%0AAny%20code%20%28including%20test%20helpers%20or%20future%20callers%29%20that%20does%20%60HILLSTROM_FROZEN_PARAMS%5B%22email_share%22%5D%20%3D%202.0%60%20or%20%60HILLSTROM_FROZEN_PARAMS.pop%28...%29%60%20will%20silently%20corrupt%20the%20frozen%20params%20injected%20by%20every%20subsequent%20%60HillstromPolicyRunner._forward_params%60%20call%20in%20that%20process%20%E2%80%94%20including%20parallel%20benchmark%20runs.%20Wrapping%20it%20in%20%60types.MappingProxyType%60%20makes%20accidental%20mutation%20an%20immediate%20%60TypeError%60%3A%0A%0A%60%60%60python%0Aimport%20types%0A%0AHILLSTROM_FROZEN_PARAMS%3A%20types.MappingProxyType%5Bstr%2C%20float%5D%20%3D%20types.MappingProxyType%28%0A%20%20%20%20%7B%0A%20%20%20%20%20%20%20%20%22email_share%22%3A%201.0%2C%0A%20%20%20%20%20%20%20%20%22social_share_of_remainder%22%3A%200.0%2C%0A%20%20%20%20%20%20%20%20%22min_propensity_clip%22%3A%200.01%2C%0A%20%20%20%20%7D%0A%29%0A%60%60%60%0A%0AThe%20%60_forward_params%60%20loop%20%28%60for%20key%2C%20value%20in%20HILLSTROM_FROZEN_PARAMS.items%28%29%60%29%20works%20unchanged%20on%20a%20%60MappingProxyType%60.%0A%0A&repo=datablogin%2Fcausal-optimizer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=2">Fix All in Claude Code</picture></a>

`````markdown
Path: causal_optimizer/benchmarks/__init__.py
Line: 23-25

Comment:
**Two import blocks from the same module**

`VALID_STRATEGIES as HILLSTROM_VALID_STRATEGIES` is split into its own block solely because it's a renamed import. Merging it into the first block is cleaner and avoids two separate module loads (Python deduplicates them, but the visual split suggests a distinction that doesn't exist at runtime):

```suggestion
from causal_optimizer.benchmarks.hillstrom import (
    HILLSTROM_FROZEN_PARAMS,
    HILLSTROM_POOLED_PROPENSITY,
    HILLSTROM_PRIMARY_PROPENSITY,
    VALID_STRATEGIES as HILLSTROM_VALID_STRATEGIES,
    HillstromBenchmarkResult,
    HillstromPolicyRunner,
    HillstromScenario,
    HillstromSliceType,
    hillstrom_active_search_space,
    hillstrom_null_baseline,
    hillstrom_projected_prior_graph,
    load_hillstrom_slice,
    permute_hillstrom_spend,
)
```

---

Path: causal_optimizer/benchmarks/hillstrom.py
Line: 168-172

Comment:
**`HILLSTROM_FROZEN_PARAMS` is a mutable module-level dict**

Any code (including test helpers or future callers) that does `HILLSTROM_FROZEN_PARAMS["email_share"] = 2.0` or `HILLSTROM_FROZEN_PARAMS.pop(...)` will silently corrupt the frozen params injected by every subsequent `HillstromPolicyRunner._forward_params` call in that process — including parallel benchmark runs. Wrapping it in `types.MappingProxyType` makes accidental mutation an immediate `TypeError`:

```python
import types

HILLSTROM_FROZEN_PARAMS: types.MappingProxyType[str, float] = types.MappingProxyType(
    {
        "email_share": 1.0,
        "social_share_of_remainder": 0.0,
        "min_propensity_clip": 0.01,
    }
)
```

The `_forward_params` loop (`for key, value in HILLSTROM_FROZEN_PARAMS.items()`) works unchanged on a `MappingProxyType`.

`````

</details>

<sub>Reviews (4): Last reviewed commit: ["fix: sanitize hillstrom\_provenance befor..."](https://github.com/datablogin/causal-optimizer/commit/bd118093429b6a4ca60234de3060a04f059df8f6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28517457)</sub>

### DIFF
--- a/causal_optimizer/benchmarks/__init__.py
+++ b/causal_optimizer/benchmarks/__init__.py
@@ -27,6 +27,23 @@ from causal_optimizer.benchmarks.dose_response import (
     evaluate_protocol,
 )
 from causal_optimizer.benchmarks.high_dimensional import HighDimensionalSparseBenchmark
+from causal_optimizer.benchmarks.hillstrom import (
+    HILLSTROM_FROZEN_PARAMS,
+    HILLSTROM_POOLED_PROPENSITY,
+    HILLSTROM_PRIMARY_PROPENSITY,
+    HillstromBenchmarkResult,
+    HillstromPolicyRunner,
+    HillstromScenario,
+    HillstromSliceType,
+    hillstrom_active_search_space,
+    hillstrom_null_baseline,
+    hillstrom_projected_prior_graph,
+    load_hillstrom_slice,
+    permute_hillstrom_spend,
+)
+from causal_optimizer.benchmarks.hillstrom import (
+    VALID_STRATEGIES as HILLSTROM_VALID_STRATEGIES,
+)
 from causal_optimizer.benchmarks.interaction import InteractionBenchmark
 from causal_optimizer.benchmarks.interaction_policy import (
     InteractionPolicyScenario,
@@ -81,6 +98,10 @@ class BenchmarkSCM(Protocol):
 
 
 __all__ = [
+    "HILLSTROM_FROZEN_PARAMS",
+    "HILLSTROM_POOLED_PROPENSITY",
+    "HILLSTROM_PRIMARY_PROPENSITY",
+    "HILLSTROM_VALID_STRATEGIES",
     "BenchmarkResult",
     "BenchmarkRunner",
     "BenchmarkSCM",
@@ -92,10 +113,14 @@ __all__ = [
     "DoseResponseScenario",
     "HighDimensionalSparseBenchmark",
     "HighNoiseDemandResponse",
+    "HillstromBenchmarkResult",
+    "HillstromPolicyRunner",
+    "HillstromScenario",
+    "HillstromSliceType",
     "InteractionBenchmark",
     "InteractionPolicyScenario",
-    "MediumNoiseDemandResponse",
     "InteractionSCM",
+    "MediumNoiseDemandResponse",
     "NullSignalResult",
     "PolicyRunner",
     "PredictiveBenchmarkResult",
@@ -108,10 +133,15 @@ __all__ = [
     "evaluate_on_test",
     "evaluate_policy",
     "evaluate_protocol",
+    "hillstrom_active_search_space",
+    "hillstrom_null_baseline",
+    "hillstrom_projected_prior_graph",
     "interaction_propensity",
     "interaction_treatment_effect",
     "load_energy_frame",
+    "load_hillstrom_slice",
     "net_benefit",
+    "permute_hillstrom_spend",
     "permute_target",
     "propensity",
     "run_null_strategy",
--- a/causal_optimizer/benchmarks/hillstrom.py
+++ b/causal_optimizer/benchmarks/hillstrom.py
@@ -1,0 +1,587 @@
+"""Sprint 31 Hillstrom benchmark harness.
+
+Implements the launch contract locked in the Sprint 31 Hillstrom
+benchmark contract doc. Reuses ``MarketingLogAdapter`` unchanged via a
+narrow loader + wrapped runner pattern; no changes to
+``causal_optimizer/domain_adapters/marketing_logs.py`` are required.
+
+Contract anchors
+----------------
+
+- Primary slice: ``Womens E-Mail`` vs ``No E-Mail`` (binary).
+- Pooled secondary slice: ``Any E-Mail`` (``Mens E-Mail`` +
+  ``Womens E-Mail``) vs ``No E-Mail`` (binary).
+- Per-slice propensity pinned as an exact constant:
+  ``0.5`` for the primary slice (computed as ``(1/3) / (2/3)``) and
+  ``2.0 / 3.0`` for the pooled slice (computed as ``2.0 / 3.0`` in
+  code, not a rounded constant).
+- Frozen parameter dimensions (pre-baked in every forwarded call):
+  ``email_share = 1.0``, ``social_share_of_remainder = 0.0``,
+  ``min_propensity_clip = 0.01``. Active tuned search space is therefore
+  3 variables: ``eligibility_threshold``, ``regularization``,
+  ``treatment_budget_pct``.
+- Prior graph: projected to the 7-edge sub-DAG over the active nodes.
+  The three frozen variables and the 7 edges incident to them are
+  dropped.
+- Null control: permuted-outcome null baseline is the raw scalar
+  ``μ = mean(spend)`` computed on the unshuffled reshaped frame; this
+  avoids the misleading "``policy_value`` under a uniform-treatment
+  policy" framing, which on randomized data only recovers the
+  treated-arm mean (see ``marketing_logs.py:322``).
+
+Public API
+----------
+- :class:`HillstromSliceType`
+- :func:`load_hillstrom_slice`
+- :func:`hillstrom_active_search_space`
+- :func:`hillstrom_projected_prior_graph`
+- :class:`HillstromPolicyRunner`
+- :func:`permute_hillstrom_spend`
+- :func:`hillstrom_null_baseline`
+- :class:`HillstromBenchmarkResult`
+- :class:`HillstromScenario`
+- :data:`HILLSTROM_FROZEN_PARAMS`
+- :data:`HILLSTROM_PRIMARY_PROPENSITY`
+- :data:`HILLSTROM_POOLED_PROPENSITY`
+"""
+
+from __future__ import annotations
+
+import time
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any
+
+import numpy as np
+import pandas as pd
+
+from causal_optimizer.benchmarks.runner import sample_random_params
+from causal_optimizer.domain_adapters.marketing_logs import MarketingLogAdapter
+from causal_optimizer.engine.loop import ExperimentEngine
+from causal_optimizer.types import (
+    CausalGraph,
+    SearchSpace,
+    Variable,
+    VariableType,
+)
+
+# ── Contract constants ──────────────────────────────────────────────
+
+HILLSTROM_PRIMARY_PROPENSITY: float = 0.5
+"""Primary slice propensity: ``(1/3) / (2/3) = 0.5`` exactly."""
+
+HILLSTROM_POOLED_PROPENSITY: float = 2.0 / 3.0
+"""Pooled slice propensity: ``(1/3 + 1/3) / 1 = 2/3`` exactly.
+
+Computed as ``2.0 / 3.0`` rather than a rounded constant like ``0.667``
+so the Sprint 31 smoke-test invariant ``propensity == 2.0 / 3.0`` is
+bitwise-true on every pooled-slice row.
+"""
+
+HILLSTROM_FROZEN_PARAMS: dict[str, float] = {
+    "email_share": 1.0,
+    "social_share_of_remainder": 0.0,
+    "min_propensity_clip": 0.01,
+}
+"""Frozen ``MarketingLogAdapter`` dimensions for Hillstrom.
+
+All three are degenerate on Hillstrom:
+
+- ``email_share`` and ``social_share_of_remainder``: Hillstrom treatment
+  is single-channel e-mail, so there is no cross-channel allocation
+  problem to solve.
+- ``min_propensity_clip``: propensity is a per-slice constant
+  (``0.5`` or ``2/3``) and both are ``>= 0.5`` (the upper bound of
+  ``min_propensity_clip``'s range), so the clip never activates and
+  the optimizer sees a flat response surface along this dimension.
+"""
+
+HILLSTROM_TREATED_ARMS: frozenset[str] = frozenset({"Mens E-Mail", "Womens E-Mail"})
+HILLSTROM_CONTROL_ARM: str = "No E-Mail"
+HILLSTROM_TREATED_COST: float = 0.10
+HILLSTROM_CONTROL_COST: float = 0.0
+
+# history_segment bucket map — matches the raw Hillstrom CSV strings,
+# numeric prefix included, per Sprint 31 contract Section 3c.
+_HISTORY_SEGMENT_TO_SEGMENT: dict[str, str] = {
+    "1) $0 - $100": "low",
+    "2) $100 - $200": "low",
+    "3) $200 - $350": "medium",
+    "4) $350 - $500": "medium",
+    "5) $500 - $750": "high_value",
+    "6) $750 - $1,000": "high_value",
+    "7) $1,000 +": "high_value",
+}
+
+
+# ── Slice type ───────────────────────────────────────────────────────
+
+
+class HillstromSliceType(str, Enum):
+    """Which Hillstrom binary slice to load.
+
+    ``PRIMARY``
+        ``Womens E-Mail`` (``treatment=1``) vs ``No E-Mail``
+        (``treatment=0``). Drops ``Mens E-Mail`` rows. Propensity is
+        exactly ``0.5``.
+    ``POOLED``
+        ``Any E-Mail`` (``Mens`` or ``Womens``, ``treatment=1``) vs
+        ``No E-Mail`` (``treatment=0``). Keeps every raw row.
+        Propensity is exactly ``2/3``.
+    """
+
+    PRIMARY = "primary"
+    POOLED = "pooled"
+
+
+# ── Loader ───────────────────────────────────────────────────────────
+
+
+def _treatment_from_segment(segment: pd.Series, slice_type: HillstromSliceType) -> pd.Series:
+    """Map the raw ``segment`` column to binary ``treatment``."""
+    if slice_type is HillstromSliceType.PRIMARY:
+        return (segment == "Womens E-Mail").astype(int)
+    return segment.isin(HILLSTROM_TREATED_ARMS).astype(int)
+
+
+def _propensity_for_slice(slice_type: HillstromSliceType) -> float:
+    if slice_type is HillstromSliceType.PRIMARY:
+        return HILLSTROM_PRIMARY_PROPENSITY
+    return HILLSTROM_POOLED_PROPENSITY
+
+
+def load_hillstrom_slice(
+    raw: pd.DataFrame,
+    *,
+    slice_type: HillstromSliceType,
+    treated_cost: float = HILLSTROM_TREATED_COST,
+    control_cost: float = HILLSTROM_CONTROL_COST,
+) -> pd.DataFrame:
+    """Reshape a raw Hillstrom frame into the ``MarketingLogAdapter`` schema.
+
+    Filters to the requested slice, maps ``segment`` → ``treatment``,
+    passes through ``spend`` → ``outcome``, assigns a constant per-slice
+    propensity, a fixed per-send cost, the constant ``channel = "email"``,
+    and a bucketed ``segment`` column derived from ``history_segment``.
+    The ``visit`` and ``conversion`` columns are retained on the output
+    frame as secondary reported outcomes (not as adapter descriptors),
+    matching Sprint 31 contract Section 4c.
+
+    Args:
+        raw: DataFrame with the canonical Hillstrom column schema.
+        slice_type: Which binary slice to produce.
+        treated_cost: Per-treated-row fixed cost. Defaults to
+            ``HILLSTROM_TREATED_COST``.
+        control_cost: Per-control-row fixed cost. Defaults to
+            ``HILLSTROM_CONTROL_COST``.
+
+    Returns:
+        A new DataFrame with the ``MarketingLogAdapter`` required
+        columns (``treatment``, ``outcome``, ``cost``) plus the optional
+        columns (``propensity``, ``channel``, ``segment``) and the
+        retained Hillstrom ``visit`` and ``conversion`` columns.
+
+    Raises:
+        ValueError: If ``raw`` is missing any required Hillstrom
+            column.
+    """
+    required = {"segment", "spend", "visit", "conversion", "history_segment"}
+    missing = required - set(raw.columns)
+    if missing:
+        msg = f"Hillstrom frame is missing required columns: {sorted(missing)}"
+        raise ValueError(msg)
+
+    frame = raw.copy()
+
+    # Slice filter
+    if slice_type is HillstromSliceType.PRIMARY:
+        keep = frame["segment"].isin({"Womens E-Mail", HILLSTROM_CONTROL_ARM})
+        frame = frame.loc[keep].reset_index(drop=True)
+    # Pooled slice keeps every row.
+
+    treatment = _treatment_from_segment(frame["segment"], slice_type)
+    propensity = _propensity_for_slice(slice_type)
+    cost = np.where(treatment == 1, treated_cost, control_cost)
+
+    # Bucket history_segment → {"low", "medium", "high_value"}. Any
+    # unknown bucket is mapped to "low" to avoid silent NaNs; this
+    # should not occur on canonical Hillstrom data but is defensive.
+    segment_bucket = frame["history_segment"].map(_HISTORY_SEGMENT_TO_SEGMENT).fillna("low")
+
+    reshaped = pd.DataFrame(
+        {
+            "treatment": treatment.astype(int),
+            "outcome": frame["spend"].astype(float),
+            "cost": cost.astype(float),
+            "propensity": np.full(len(frame), propensity, dtype=float),
+            "channel": "email",
+            "segment": segment_bucket.astype(str),
+            # Retained Hillstrom outcomes as secondary reported outcomes.
+            "visit": frame["visit"].astype(int),
+            "conversion": frame["conversion"].astype(int),
+        }
+    )
+
+    return reshaped
+
+
+# ── Active search space ──────────────────────────────────────────────
+
+
+_ACTIVE_VAR_NAMES: tuple[str, ...] = (
+    "eligibility_threshold",
+    "regularization",
+    "treatment_budget_pct",
+)
+
+
+def hillstrom_active_search_space() -> SearchSpace:
+    """Return the 3-variable active search space for Hillstrom.
+
+    Bounds are inherited from ``MarketingLogAdapter.get_search_space()``
+    so the active subspace is numerically identical to the adapter's
+    native ranges for the three tuned dimensions. The three frozen
+    dimensions (``email_share``, ``social_share_of_remainder``,
+    ``min_propensity_clip``) are pre-baked by
+    :class:`HillstromPolicyRunner` and do not appear in this space.
+    """
+    # Build a throwaway adapter against a minimal placeholder frame to
+    # extract the canonical bounds. Note: the adapter's __init__ requires
+    # non-empty data with treatment/outcome/cost columns, so we use a
+    # 2-row stub. This is not a data dependency — it only reads the
+    # adapter's static variable bounds.
+    stub = pd.DataFrame(
+        {
+            "treatment": [0, 1],
+            "outcome": [0.0, 1.0],
+            "cost": [0.0, 0.0],
+        }
+    )
+    full = MarketingLogAdapter(data=stub).get_search_space()
+    by_name = {v.name: v for v in full.variables}
+    variables = [
+        Variable(
+            name=name,
+            variable_type=VariableType.CONTINUOUS,
+            lower=by_name[name].lower,
+            upper=by_name[name].upper,
+        )
+        for name in _ACTIVE_VAR_NAMES
+    ]
+    return SearchSpace(variables=variables)
+
+
+# ── Projected prior graph ────────────────────────────────────────────
+
+
+def hillstrom_projected_prior_graph() -> CausalGraph:
+    """Return the 7-edge projected prior graph for Hillstrom.
+
+    The projection drops the 3 frozen variables and the 7 edges
+    incident to them from ``MarketingLogAdapter.get_prior_graph()``,
+    leaving a sub-DAG over the active nodes. See Sprint 31 contract
+    Section 4a.i for the enumeration and rationale.
+    """
+    return CausalGraph(
+        edges=[
+            ("eligibility_threshold", "treated_fraction"),
+            ("treatment_budget_pct", "treated_fraction"),
+            ("regularization", "treated_fraction"),
+            ("regularization", "policy_value"),
+            ("treated_fraction", "total_cost"),
+            ("treated_fraction", "policy_value"),
+            ("treated_fraction", "effective_sample_size"),
+        ],
+    )
+
+
+# ── Wrapped runner ───────────────────────────────────────────────────
+
+
+class HillstromPolicyRunner:
+    """ExperimentRunner that pre-bakes frozen params before adapter call.
+
+    Accepts an active-only parameter dict (with the three tuned
+    variables) and forwards to ``MarketingLogAdapter.run_experiment``
+    after injecting the three frozen constants in
+    :data:`HILLSTROM_FROZEN_PARAMS`. Does not mutate the caller's dict
+    and does not modify the adapter.
+
+    The returned metrics include ``MarketingLogAdapter``'s full
+    per-run dict (``policy_value``, ``total_cost``, ``treated_fraction``,
+    ``effective_sample_size``, ``propensity_clip_fraction``,
+    ``max_ips_weight``, ``weight_cv``, ``zero_support``) plus an
+    ``objective`` alias equal to the negated ``policy_value`` — because
+    ``policy_value`` is maximized but the engine minimizes ``objective``.
+    """
+
+    def __init__(self, adapter: MarketingLogAdapter) -> None:
+        self._adapter = adapter
+
+    def _forward_params(self, active_params: dict[str, Any]) -> dict[str, Any]:
+        """Return a new dict with frozen params injected.
+
+        Does not mutate the caller's ``active_params`` dict.
+        """
+        forwarded: dict[str, Any] = dict(active_params)
+        for key, value in HILLSTROM_FROZEN_PARAMS.items():
+            forwarded[key] = value
+        return forwarded
+
+    def run(self, parameters: dict[str, Any]) -> dict[str, float]:
+        """Evaluate one active-only parameter dict on the wrapped adapter."""
+        forwarded = self._forward_params(parameters)
+        metrics = self._adapter.run_experiment(forwarded)
+        # Engine minimizes "objective"; Hillstrom maximizes policy_value.
+        metrics["objective"] = -metrics["policy_value"]
+        return metrics
+
+
+# ── Null control helpers ─────────────────────────────────────────────
+
+
+def permute_hillstrom_spend(
+    reshaped: pd.DataFrame, *, seed: int, outcome_col: str = "outcome"
+) -> pd.DataFrame:
+    """Return a copy of ``reshaped`` with the ``outcome`` column permuted.
+
+    The permutation is deterministic in ``seed`` and preserves:
+
+    - the multiset of ``outcome`` values (permutation of the same column
+      — so ``μ = mean(outcome)`` is identical on the original and the
+      shuffled frame)
+    - every other column, including ``treatment``, ``propensity``,
+      ``cost``, ``channel``, ``segment``, ``visit``, and ``conversion``.
+
+    Args:
+        reshaped: A frame already reshaped by :func:`load_hillstrom_slice`.
+        seed: RNG seed. Use one seed per null-control replicate.
+        outcome_col: Name of the column to permute. Defaults to
+            ``"outcome"``.
+
+    Returns:
+        A new frame with the outcome column shuffled; the original
+        frame is not modified.
+
+    Raises:
+        ValueError: If ``outcome_col`` is not in the frame.
+    """
+    if outcome_col not in reshaped.columns:
+        raise ValueError(f"column {outcome_col!r} not in reshaped frame")
+    out = reshaped.copy()
+    rng = np.random.default_rng(seed)
+    values = out[outcome_col].to_numpy().copy()
+    rng.shuffle(values)
+    out[outcome_col] = values
+    return out
+
+
+def hillstrom_null_baseline(reshaped: pd.DataFrame, *, outcome_col: str = "outcome") -> float:
+    """Return the Sprint 31 null-control baseline ``μ = mean(outcome)``.
+
+    Called once on the original (unshuffled) reshaped frame; shuffling
+    is a permutation and does not change the column mean, so the same
+    scalar is used for every null-control seed.
+    """
+    return float(reshaped[outcome_col].astype(float).mean())
+
+
+# ── Benchmark scenario ───────────────────────────────────────────────
+
+
+_VALID_STRATEGIES: frozenset[str] = frozenset({"random", "surrogate_only", "causal"})
+
+
+@dataclass
+class HillstromBenchmarkResult:
+    """Result of running one strategy on one Hillstrom slice.
+
+    Attributes:
+        strategy: ``"random"`` / ``"surrogate_only"`` / ``"causal"``.
+        budget: Experiments run by the strategy.
+        seed: RNG seed for the optimizer.
+        slice_type: ``"primary"`` or ``"pooled"``.
+        is_null_control: ``True`` when evaluated on a permuted-outcome
+            frame (Sprint 31 null-control path).
+        policy_value: Best ``policy_value`` found by the strategy on the
+            given slice (higher is better).
+        selected_parameters: Best active-only parameter dict, or
+            ``None`` if no valid result was produced.
+        runtime_seconds: Wall-clock time for the full strategy run.
+    """
+
+    strategy: str
+    budget: int
+    seed: int
+    slice_type: str
+    is_null_control: bool
+    policy_value: float
+    selected_parameters: dict[str, Any] | None = None
+    runtime_seconds: float = 0.0
+    null_baseline: float | None = None
+    # Secondary reported outcomes — IPS-unweighted per-strategy aggregates
+    # taken over the logged observations matched by the best policy.
+    secondary_outcomes: dict[str, float] = field(default_factory=dict)
+
+
+class HillstromScenario:
+    """Top-level Hillstrom benchmark scenario.
+
+    Wraps a reshaped Hillstrom frame in a :class:`MarketingLogAdapter`
+    and a :class:`HillstromPolicyRunner`, then runs one strategy at a
+    given ``(budget, seed)`` pair. Supports both the real slice and
+    the permuted-outcome null-control path.
+
+    Note: the scenario is intended as the unit of work for the Sprint 31
+    benchmark runner; a separate CLI script composes scenarios over a
+    grid of (strategy, budget, seed) combinations.
+    """
+
+    def __init__(
+        self,
+        raw: pd.DataFrame,
+        *,
+        slice_type: HillstromSliceType,
+    ) -> None:
+        self._raw = raw
+        self._slice_type = slice_type
+        self._real_slice = load_hillstrom_slice(raw, slice_type=slice_type)
+        # The null-baseline μ is a scalar computed once from the real
+        # (unshuffled) frame; shuffling preserves the column mean.
+        self._null_baseline = hillstrom_null_baseline(self._real_slice)
+
+    @property
+    def slice_type(self) -> HillstromSliceType:
+        return self._slice_type
+
+    @property
+    def null_baseline(self) -> float:
+        return self._null_baseline
+
+    @property
+    def real_slice(self) -> pd.DataFrame:
+        return self._real_slice
+
+    def run_strategy(
+        self,
+        strategy: str,
+        *,
+        budget: int,
+        seed: int,
+        null_control: bool = False,
+    ) -> HillstromBenchmarkResult:
+        """Run one strategy on this scenario and return the result.
+
+        Args:
+            strategy: One of ``"random"``, ``"surrogate_only"``,
+                ``"causal"``.
+            budget: Number of experiments (adapter evaluations).
+            seed: Optimizer RNG seed.
+            null_control: When ``True``, the strategy runs against a
+                permuted-outcome copy of the reshaped frame. Permutation
+                seed is the same ``seed`` argument so null-control seeds
+                are reproducible.
+
+        Returns:
+            A :class:`HillstromBenchmarkResult`.
+
+        Raises:
+            ValueError: If ``strategy`` is not in ``_VALID_STRATEGIES``.
+        """
+        if strategy not in _VALID_STRATEGIES:
+            msg = f"Unknown strategy {strategy!r}. Must be one of {sorted(_VALID_STRATEGIES)}."
+            raise ValueError(msg)
+
+        t_start = time.perf_counter()
+
+        frame = (
+            permute_hillstrom_spend(self._real_slice, seed=seed)
+            if null_control
+            else self._real_slice
+        )
+        adapter = MarketingLogAdapter(data=frame, seed=seed)
+        runner = HillstromPolicyRunner(adapter=adapter)
+        space = hillstrom_active_search_space()
+
+        best_policy_value = float("-inf")
+        best_params: dict[str, Any] | None = None
+
+        if strategy == "random":
+            rng = np.random.default_rng(seed)
+            for _ in range(budget):
+                params = sample_random_params(space, rng)
+                metrics = runner.run(params)
+                pv = metrics["policy_value"]
+                if pv > best_policy_value:
+                    best_policy_value = pv
+                    best_params = params
+        else:
+            graph = hillstrom_projected_prior_graph() if strategy == "causal" else None
+            engine = ExperimentEngine(
+                search_space=space,
+                runner=runner,
+                causal_graph=graph,
+                objective_name="objective",
+                minimize=True,
+                seed=seed,
+            )
+            engine.run_loop(budget)
+            best_result = engine.log.best_result("objective", minimize=True)
+            if best_result is not None:
+                best_params = best_result.parameters
+                best_policy_value = best_result.metrics.get("policy_value", float("-inf"))
+
+        runtime = time.perf_counter() - t_start
+
+        secondary: dict[str, float] = {}
+        if best_params is not None and not null_control:
+            # Secondary outcomes: IPS-unweighted treated-arm means of the
+            # retained Hillstrom columns, taken over the rows actually
+            # selected by the best policy. This is a report-time
+            # diagnostic, not an optimization target.
+            secondary = _secondary_outcomes_under_policy(frame, best_params)
+
+        return HillstromBenchmarkResult(
+            strategy=strategy,
+            budget=budget,
+            seed=seed,
+            slice_type=self._slice_type.value,
+            is_null_control=null_control,
+            policy_value=(
+                best_policy_value if best_policy_value != float("-inf") else float("nan")
+            ),
+            selected_parameters=best_params,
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the codebase. Here's my review:

---

## PR Review: Sprint 31 Hillstrom Benchmark Harness

### Overall Assessment

This is a well-structured, carefully designed PR. The "loader + wrapped runner" pattern cleanly integrates a new domain benchmark without modifying the existing `MarketingLogAdapter`. The code is thoroughly documented and the test coverage is excellent. Below are my findings, organized by severity.

---

### Issues

**1. `hillstrom_active_search_space()` creates a throwaway adapter on every call** (hillstrom.py:248-260)

This function instantiates a `MarketingLogAdapter` with a stub DataFrame just to read static variable bounds. It's called inside `run_strategy()` (line 510), meaning it runs once per `(strategy, budget, seed)` combination. The bounds are constants that never change.

Consider caching the result at module level or using `functools.lru_cache()`. Not a correctness bug, but a repeated allocation that's easy to avoid.

**2. `_secondary_outcomes_under_policy` doesn't use the policy's parameters** (hillstrom.py:567-594)

The function name says "under policy" but it computes simple treated/control arm means on the full reshaped frame regardless of the best parameters found. The docstring correctly documents this, and the comment at the call site (line 544-548) explains the deferral, but the function name is misleading. Consider renaming to `_secondary_outcomes_population_level` or `_secondary_arm_aggregates` to avoid confusion in Sprint 32+ when policy-conditioned outcomes arrive.

**3. `best_result.parameters` includes frozen params from the engine log but `selected_parameters` should be active-only** (hillstrom.py:537)

When using the `surrogate_only` or `causal` strategy, `best_result.parameters` comes from `engine.log.best_result()`. The engine calls `runner.run(parameters)` where `parameters` are the engine-suggested active-only params. The engine logs whatever parameters it suggested (the 3 active ones), so this is correct — the engine never sees frozen params. But this is a subtle correctness dependency that deserves a comment or assertion. If the engine ever added pass-through of runner-returned metadata back into the logged parameters, the invariant would break silently.

**4. Propensity docstring math is slightly off** (hillstrom.py:71)

The docstring says "Primary slice propensity: `(1/3) / (2/3) = 0.5`". This is the ratio of treatment probability to total (control+treatment) probability, which is indeed `(1/3) / (2/3) = 0.5` — but that's a confusing way to express it. The actual propensity (probability of receiving treatment) is `1/3 / (1/3 + 1/3) = 0.5` in the binary primary slice. The `(1/3) / (2/3)` expression is `P(Womens) / P(Womens + Control)` = `(1/3) / (2/3)` = `0.5`, which is correct but reads as "treatment odds" rather than "treatment probability". Minor docstring clarity issue.

**5. No test for the CLI script `parse_args`/`main`** (scripts/hillstrom_benchmark.py)

The script has validation functions (`_validate_budgets`, `_validate_strategies`, `_validate_slices`, `_parse_int_list`) and a `_sanitize_for_json` helper that are untested. The `_sanitize_for_json` function handles edge cases (NaN, Inf, numpy types) that are worth unit-testing. The script is covered by `pragma: no cover` on the exception handlers but the core logic paths have no tests.

**6. `_print_summary` header length mismatch with non-ASCII** (scripts/hillstrom_benchmark.py:184-185)

The header string contains `μ` (Greek mu), but the separator line uses `"-" * len(header)`. `len()` counts characters, not display width — `μ` is a single character so the count is correct. However, if the terminal font renders `μ` wider than a single ASCII character (unlikely in modern monospace terminals but possible), the alignment could drift. This is cosmetic only.

**7. `sys.exit(1)` in validation helpers makes them untestable** (scripts/hillstrom_benchmark.py:131-161)

The validation functions call `sys.exit(1)` directly instead of raising exceptions. This is a CLI pattern choice, but it means these can't be tested without `pytest.raises(SystemExit)` and the error messages go to stderr rather than being catchable. For a CLI script this is fine, but it's worth noting if you plan to reuse these validators elsewhere.

---

### Positive Observations

**Architecture**: The "narrow loader + wrapped runner" pattern is clean. `HillstromPolicyRunner` correctly satisfies the `ExperimentRunner` protocol (has `run(parameters) -> dict[str, float]`), and the frozen-param injection is non-mutating. This is the right way to adapt an existing domain adapter to a new dataset without modifying it.

**Contract enforcement through tests**: The test suite is organized around contract invariants (propensity pinning, graph projection, search space bounds, frozen params). Each invariant has a dedicated test class, making it easy to trace failures back to contract violations.

**Null control design**: The permutation-based null control is statistically sound. Using the same seed for permutation and optimizer ensures reproducibility. The `hillstrom_null_baseline` is correctly defined as `mean(outcome)` on the unshuffled frame, and the tests verify that shuffling preserves this invariant.

**Defensive validation**: `load_hillstrom_slice` validates required columns. `_HISTORY_SEGMENT_TO_SEGMENT` maps with `.fillna("low")` to handle unexpected segments. The `permute_hillstrom_spend` validates that the outcome column exists.

**JSON serialization**: `_sanitize_for_json` handles numpy types, NaN, and Inf correctly, and the `json.dump` uses `allow_nan=False` as a safety net.

**Test coverage breadth**: 
- Unit tests cover slice mapping, propensity invariants, graph projection, search space, frozen params, runner injection, null control permutation, and defensive guards.
- Integration tests cover all three strategies (random, surrogate_only, causal), both slices, null control, and secondary outcomes.
- The test that verifies `projected.issubset(full)` and `len(full - projected) == 7` is an excellent cross-check against the adapter's full graph.

**Data fixture**: The 300-row fixture is large enough to exercise the adapter's IPS logic meaningfully while small enough for fast CI. The fixture contains realistic Hillstrom column values with the expected segment distribution.

---

### Minor Nits

- `hillstrom.py:591` — line is 96 chars, within the 100-char limit but close. The parallel line 594 has the same pattern. Both are fine.
- `hillstrom.py:99` — `HILLSTROM_TREATED_ARMS` uses `frozenset` with `str` members; this is correct and the `HillstromSliceType(str, Enum)` pattern is clean.
- The integration test docstring says "3 experiments per strategy" but the tests use `budget=8`. Minor doc drift, not a code issue.

---

### Summary

This is a well-executed PR. The main harness module is clean, well-documented, and correctly integrates with the existing adapter and engine patterns. The test coverage is thorough and organized around contract invariants. The two substantive suggestions are: (1) cache the active search space to avoid repeated adapter instantiation, and (2) rename `_secondary_outcomes_under_policy` to avoid name/behavior confusion when policy-conditioned outcomes are added later. Everything else is cosmetic or minor.

---

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

Here's my review of the PR:

---

## PR Review: Sprint 31 Hillstrom Benchmark Harness

### Overall Assessment

This is a well-structured, well-documented addition. The "loader + wrapped runner" pattern is a clean way to reuse `MarketingLogAdapter` without modifying it. The contract-driven design with pinned constants, frozen dimensions, and projected graph is sound. The test coverage is thorough. A few issues worth addressing:

---

### Issues

#### 1. `assert` in library code (hillstrom.py:566) — Medium

```python
assert set(best_params) == set(_ACTIVE_VAR_NAMES), (...)
```

This is the only bare `assert` in the new code and it guards a contract invariant in production code. `assert` statements are stripped when Python runs with `-O` / `PYTHONOPTIMIZE=1`, silently removing this safety check. The existing `assert` usages in the codebase are either type-narrowing (`assert data is not None`) or internal-only (`suggest.py`). This one guards a public API contract boundary.

**Suggestion:** Replace with a `RuntimeError` or `ValueError` raise — the comment already calls it "an invariant that guards the contract boundary."

#### 2. Missing `benchmarks/__init__.py` re-exports

Every other benchmark module (`counterfactual_energy`, `dose_response`, `predictive_energy`, `interaction_policy`, etc.) has its public API re-exported from `causal_optimizer/benchmarks/__init__.py` and listed in `__all__`. The Hillstrom module has no imports or `__all__` entries there. This means `from causal_optimizer.benchmarks import HillstromScenario` won't work, breaking the pattern established by the other benchmarks.

#### 3. `_print_summary` header-width mismatch with unicode `μ`

`hillstrom_benchmark.py:187`: The separator line length is computed as `len(header)`, but the header contains `μ` (a non-ASCII character). `len()` counts codepoints, not display width. In most terminals `μ` renders at width 1, so this is fine — but it's worth noting for environments that render it at width 2. This is very minor.

#### 4. `_secondary_arm_aggregates` computes frame-level rates, not policy-conditioned ones

The function is called inside `run_strategy` and its output is attached to the `HillstromBenchmarkResult`, but it computes arm-level means over the *entire* reshaped frame regardless of which policy was selected. The code and docstring are honest about this ("deferred to a later sprint"), but the field name `secondary_outcomes` on the result dataclass might mislead readers into thinking these are policy-specific. This is documented — just flagging it for awareness.

#### 5. `lru_cache` on `hillstrom_active_search_space` returns a mutable `SearchSpace`

The `@lru_cache(maxsize=1)` means every caller gets the *same* `SearchSpace` object. If any downstream code mutates `space.variables` (e.g., `space.variables.append(...)`), it would corrupt the cached instance for all subsequent callers. Pydantic models are not frozen by default. This is safe *today* because nothing mutates it, but it's a latent hazard.

**Suggestion:** Either add `model_config = ConfigDict(frozen=True)` to `SearchSpace` (if that's a broader change you want) or document the no-mutation contract in the function docstring.

#### 6. Null-control runs always use primary slice — `--slices pooled` with `--null-control` silently skips

`hillstrom_benchmark.py:284`: `if args.null_control and "primary" in scenarios:` — if the user runs with `--slices pooled --null-control`, the null-control section is silently skipped because `"primary"` is not in `scenarios`. This matches the Sprint 31 contract ("primary slice only"), but a `logger.info` message when `--null-control` is set but `primary` is not in slices would prevent user confusion.

#### 7. `HillstromBenchmarkResult` uses `dataclass` while other result types use a mix

Looking at the codebase, `CounterfactualBenchmarkResult`, `DoseResponseBenchmarkResult`, `PredictiveBenchmarkResult` are all `@dataclass`. So the choice here is consistent. The `field(default_factory=dict)` for `secondary_outcomes` is correct for a mutable default.

---

### Minor / Stylistic

- **`_forward_params` loop vs dict merge**: `{**active_params, **HILLSTROM_FROZEN_PARAMS}` would be more concise than the explicit loop, but the current approach is clear and the comment about non-mutation is valuable. Fine as-is.

- **`_HISTORY_SEGMENT_TO_SEGMENT` defensive `.fillna("low")`**: Good defensive coding. Worth a comment that this should never fire on canonical data (already present).

- **`permute_hillstrom_spend` copies twice**: `out = reshaped.copy()` then `values = out[outcome_col].to_numpy().copy()`. The second `.copy()` before `rng.shuffle(values)` is necessary because `.to_numpy()` may return a view. Correct.

- **`_load_raw` in the script**: Only handles `.parquet` and `.csv`. If someone passes `.tsv` or `.xlsx`, it falls through to `pd.read_csv` which may or may not work. Acceptable for a benchmark script.

---

### Test Coverage

The test suite is strong:
- Unit tests cover all contract invariants (slice mapping, propensity pins, graph projection, frozen params, search space bounds, null-control permutation).
- Integration tests exercise the full `HillstromScenario.run_strategy` path for all three strategies.
- Script tests cover CLI parsing, validation helpers, and the JSON sanitizer.
- Edge cases: missing columns, unknown strategies/slices, empty secondary outcomes.

**One gap**: There's no test that `HillstromScenario.run_strategy` with the engine path (`surrogate_only` or `causal`) actually respects the `seed` determinism guarantee. The integration test `test_null_control_determinism_under_fixed_seed` only tests the `random` strategy path. The engine path seeds may or may not be fully deterministic depending on the backend (Ax vs RF fallback).

---

### Summary

Strong PR. The main actionable item is replacing the `assert` with a proper exception (item 1) and adding the `__init__.py` re-exports (item 2). The rest are minor observations. The code is well-documented, the test coverage is comprehensive, and the pattern is consistent with the other benchmark modules.

---

datablogin:
# Claude Code Review

## Review Feedback

Good. I now have a comprehensive understanding of the PR. Here's my review:

---

## PR Review: Sprint 31 Hillstrom Benchmark Harness

**Overall assessment: Well-structured PR.** The "loader + wrapped runner" pattern is a good architectural choice that avoids modifying `MarketingLogAdapter`. The contract-driven approach with pinned invariants and comprehensive test coverage is strong. A few issues to address:

---

### Issues

**1. `HillstromScenario._raw` is stored but never read** (`hillstrom.py:476`)

`self._raw = raw` is set in `__init__` but never referenced again. The only use of `raw` is to build `self._real_slice` via `load_hillstrom_slice()`. This is dead state that holds a potentially large DataFrame in memory for the lifetime of the scenario.

Either remove `self._raw = raw` or, if it's intended for future use, add a comment noting that.

**2. `float("-inf") != float("-inf")` comparison is fragile** (`hillstrom.py:591-593`)

```python
policy_value=(
    best_policy_value if best_policy_value != float("-inf") else float("nan")
),
```

This works in CPython (`float("-inf") == float("-inf")` is `True` per IEEE 754), but using `math.isinf()` would be more idiomatic and clearer in intent:

```python
policy_value=best_policy_value if math.isfinite(best_policy_value) else float("nan")
```

This also handles `+inf` which could theoretically arise from a buggy adapter.

**3. `_secondary_arm_aggregates` is always called on the same frame regardless of what the optimizer found** (`hillstrom.py:576-583`)

The secondary outcomes are computed over the entire reshaped frame's treated/control arms, not filtered by the best policy's selections. The code comments acknowledge this ("deferred to a later sprint"), so this is a documentation/correctness observation rather than a bug — but the variable name `secondary_outcomes` in `HillstromBenchmarkResult` may mislead downstream readers into thinking these are policy-conditioned.

**4. `_fmt_mean_std` uses population std, not sample std** (`hillstrom_benchmark.py:172`)

```python
return f"{arr.mean():.4f} +/- {arr.std():.4f}"
```

`np.ndarray.std()` defaults to `ddof=0` (population std). Per your MEMORY.md convention: "population std (ddof=0) in tables, sample-pooled (ddof=1) for Cohen's d." This is a summary table, so population std is consistent with your convention. Just confirming this is intentional.

**5. `lru_cache` on `hillstrom_active_search_space` and test interaction** (`hillstrom.py:247`)

The `lru_cache(maxsize=1)` means the cached `SearchSpace` is shared across all test cases in a process. The docstring correctly warns "do not mutate the returned object," but `SearchSpace` is a Pydantic `BaseModel` — it's not frozen by default. If any future test or code mutates the returned space's `variables` list (e.g., `space.variables.append(...)`), it would corrupt all subsequent callers silently. Consider adding `model_config = ConfigDict(frozen=True)` to `SearchSpace` if feasible, or at minimum note this risk.

**6. No test for `_load_raw` Parquet path** (`hillstrom_benchmark.py:203-208`)

`_load_raw` supports `.parquet` files but there's no test exercising that branch. Minor gap — the CSV path is well-tested.

**7. Integration tests are not marked `slow`** (`test_hillstrom_benchmark.py`)

The integration tests call `scenario.run_strategy()` which runs the full engine loop (8 experiments each for random, surrogate_only, causal). Looking at the project convention (`uv run pytest -m "not slow"` for fast tests), these integration tests don't have a `@pytest.mark.slow` marker. If they run in <5s each they're probably fine in the fast suite, but worth confirming — the surrogate_only and causal strategies construct `ExperimentEngine` and run full optimization loops.

### Nits / Minor

**8. Duplicate import block in `__init__.py`** (lines 30-46)

The `VALID_STRATEGIES` re-export uses a separate `from causal_optimizer.benchmarks.hillstrom import` block:

```python
from causal_optimizer.benchmarks.hillstrom import (
    VALID_STRATEGIES as HILLSTROM_VALID_STRATEGIES,
)
```

This could be combined into the first import block. It's functionally fine but looks like an artifact of iterative editing.

**9. `_print_summary` header line length depends on Unicode** (`hillstrom_benchmark.py:184`)

The `μ (baseline)` header contains a Unicode character. The `"-" * len(header)` separator will be character-count-based not display-width-based. This is cosmetic and only matters in terminals — not a real issue.

**10. Script test uses `sys.path` manipulation** (`test_hillstrom_benchmark_script.py:22-24`)

```python
_SCRIPTS_DIR = str(Path(__file__).resolve().parent.parent.parent / "scripts")
if _SCRIPTS_DIR not in sys.path:
    sys.path.insert(0, _SCRIPTS_DIR)
```

The comment says this follows the existing `conftest.py` pattern for `energy_predictive_benchmark.py`, so it's consistent with the codebase. But `sys.path` manipulation is a process-global side effect that persists across tests. Fine for now since this is an established pattern.

### What's Good

- **Contract-driven invariants.** The propensity constants, frozen params, and active search space are all tested as exact equalities against the Sprint 31 contract. The `_check_active_params_invariant` guard is a smart defensive measure.
- **Shared source of truth.** `VALID_STRATEGIES` is imported by the CLI rather than duplicated, with a regression test (`test_cli_imports_from_benchmark_module`) to enforce it.
- **Null control design.** The permutation-based null uses `np.random.default_rng(seed)` for reproducibility, and the test suite validates determinism, multiset preservation, and column isolation.
- **JSON sanitizer.** Thorough handling of numpy types, NaN, and inf. Well-tested.
- **Test coverage is comprehensive.** Unit tests cover slice mapping, propensity invariants, graph projection, search space, runner forwarding, null control, defensive guards, and the active-params invariant. Integration tests cover all three strategies plus pooled/primary/null-control paths.
- **`HillstromPolicyRunner._forward_params` does not mutate the caller's dict.** Correctly uses `dict(active_params)` to create a copy. Tested.

---

datablogin:
# Claude Code Review

## Review Feedback

Here's my review of PR #168: Sprint 31 Hillstrom benchmark harness.

---

## Overall Assessment

This is a well-structured, thoroughly tested PR. The code follows existing project patterns cleanly and the contract invariants are pinned by tests. I found no blocking issues, but have several observations and minor suggestions.

---

## Issues Found

### 1. `lru_cache` on function returning mutable object (Low severity)

`hillstrom.py:248` — `hillstrom_active_search_space()` caches a `SearchSpace` which contains a mutable `list[Variable]`. The docstring warns against mutation, but nothing enforces it. A caller doing `space.variables.append(...)` would corrupt all future callers silently.

This is a known pattern in the codebase and the docstring warning is clear, so it's not blocking — but worth noting for future hardening. If it ever becomes an issue, returning a frozen copy or adding `__hash__`/`__eq__` to SearchSpace would help.

### 2. No `KeyError` guard on adapter variable lookup (Low severity)

`hillstrom.py:286-293` — The list comprehension `by_name[name]` for each `_ACTIVE_VAR_NAMES` would raise an opaque `KeyError` if `MarketingLogAdapter.get_search_space()` ever renames or removes a variable. A defensive check with a clearer error message would improve debuggability:

```python
for name in _ACTIVE_VAR_NAMES:
    if name not in by_name:
        raise RuntimeError(f"MarketingLogAdapter.get_search_space() missing expected variable {name!r}")
```

### 3. No empty-frame guard after slice filter (Informational)

`hillstrom.py:208-210` — If the primary slice filter drops all rows (e.g., a CSV with no "Womens E-Mail" or "No E-Mail" rows), the code continues with an empty DataFrame. `MarketingLogAdapter.__init__` would catch this downstream, but the error message would be confusing. This can't happen on real Hillstrom data, so it's informational only.

### 4. `_secondary_arm_aggregates` called on every non-null run regardless of `best_params` (Informational)

`hillstrom.py:582-586` — The secondary outcomes are computed even when `best_params is None` is false but `null_control` is true. Looking more closely, the condition is `if best_params is not None and not null_control`, which is correct. No issue here.

### 5. Script `sys.path` manipulation in test file (Style)

`test_hillstrom_benchmark_script.py:17-19` — The `sys.path.insert(0, _SCRIPTS_DIR)` pattern is used to import the script module. This follows the existing pattern from the energy benchmark tests, so it's consistent, but worth noting that this is fragile if tests are ever run with different working directories.

---

## Code Quality Observations

**Strengths:**
- Contract fidelity is excellent — every invariant from the Sprint 31 contract is implemented and tested
- `_check_active_params_invariant()` uses `RuntimeError` instead of `assert`, surviving `PYTHONOPTIMIZE=1` — this is exactly right for a contract guard
- The "loader + wrapped runner" pattern keeps `MarketingLogAdapter` completely unchanged
- Propensity is computed as `2.0 / 3.0` (not `0.667`), preserving bitwise equality in tests
- `_sanitize_for_json` handles numpy types, NaN, and inf correctly, and `json.dump(..., allow_nan=False)` provides a belt-and-suspenders safety net
- 63 total tests across unit, script, and integration layers

**Projected prior graph correctness:**
- The 7-edge sub-DAG is verified by `test_full_adapter_graph_has_all_seven_projected_edges` which confirms the projection is a strict subset of the full adapter graph and that exactly 7 edges were dropped (14 - 7 = 7). This is a good structural invariant test.

**Null control design:**
- Permutation preserves the outcome multiset (tested), preserves all other columns (tested), and is deterministic under seed (tested). The baseline `μ = mean(spend)` is correctly computed once from the real frame and shared across null-control seeds since permutation doesn't change the column mean. Sound design.

---

## Test Coverage

Coverage is strong. Key gaps (non-blocking):

1. **No end-to-end CLI test** — The `main()` function of `scripts/hillstrom_benchmark.py` is never called in tests. The individual helpers are well-tested, and integration tests exercise the scenario layer, so the untested path is mostly I/O orchestration. Still, a small fixture-backed smoke test calling `main()` with `--data-path tests/fixtures/hillstrom_fixture.csv --budgets 3 --seeds 0 --strategies random` would close this gap.

2. **Pooled slice not tested with surrogate_only or causal** — Integration tests run `surrogate_only` and `causal` on the primary slice but only `random` on pooled. If the pooled propensity (2/3) caused any adapter issues, they'd be missed. Low risk since the adapter is propensity-agnostic at the engine level.

---

## Performance

No concerns. The `lru_cache` on `hillstrom_active_search_space()` avoids redundant adapter construction across the `(strategy, budget, seed)` grid. `permute_hillstrom_spend` is O(n) with a single copy. The secondary outcomes computation is a simple groupby-like mean, negligible cost.

---

## Summary

| Category | Verdict |
|---|---|
| Bugs | None found |
| Code quality | Excellent — clean patterns, clear contracts |
| Type safety | Good — minor `KeyError` risk on adapter bounds lookup |
| Test coverage | Strong (63 tests), minor gap on CLI e2e |
| Performance | No concerns |
| Contract fidelity | Perfect alignment |
| Documentation | Thorough docstrings and inline comments |

**Recommendation:** Merge-ready after the standard review gauntlet (claude-review.sh + greploop + human approval). The two low-severity items (KeyError guard, lru_cache mutability) are optional hardening, not blockers.

---

datablogin:
# Claude Code Review

## Review Feedback

Good, I have enough context now. Let me write the review.

---

## PR Review: Sprint 31 Hillstrom Benchmark Harness

### Overall Assessment

This is a well-structured PR that follows the existing benchmark patterns closely. The "loader + wrapped runner" pattern avoids touching `MarketingLogAdapter` and cleanly separates Hillstrom-specific concerns. The code is thorough, well-documented, and the test coverage is excellent. Below are the issues I found, grouped by severity.

---

### Issues

#### 1. Timer inconsistency: `time.perf_counter` vs `time.monotonic` (minor)

**`hillstrom.py:540`** uses `time.perf_counter()`, but the majority of the existing benchmark scenario classes (`counterfactual_energy.py`, `interaction_policy.py`, `dose_response.py`) use `time.monotonic()`. The convention within the benchmark modules isn't fully settled (newer files like `null_predictive_energy.py` also use `perf_counter`), but since `HillstromScenario.run_strategy` is the closest analog to `DoseResponseScenario.run_benchmark`, using `time.monotonic()` would be more consistent with the established pattern.

Both are fine functionally — `perf_counter` includes sleep time and has higher resolution; `monotonic` is guaranteed non-decreasing. For wall-clock benchmark timing either works. This is a style-consistency nit, not a bug.

#### 2. `lru_cache` on `hillstrom_active_search_space` returns a mutable `SearchSpace` (design concern)

**`hillstrom.py:248-270`**: The docstring correctly warns "Do not mutate the returned object", but there's no enforcement. `SearchSpace` is a Pydantic `BaseModel` — its `variables` list is a plain Python `list`, so callers *can* accidentally mutate it. The `lru_cache` then serves a corrupted copy to all subsequent callers.

This is not a bug today since no code path mutates it, but it's a latent footgun. The current design with the docstring warning is pragmatically fine, but consider either:
- Returning a `deepcopy` inside the function (defeats caching benefit slightly)
- Making the cache private and wrapping it with a public function that returns a copy
- Using `frozen=True` on SearchSpace if Pydantic supports it in v2 (it does with `model_config`)

Low priority — the docstring is adequate for now.

#### 3. `_secondary_arm_aggregates` is called unconditionally on non-null runs but the `frame` variable it receives may differ from `self._real_slice` (correctness, minor)

**`hillstrom.py:593-597`**: `_secondary_arm_aggregates(frame)` is called with `frame`, which is either `self._real_slice` (for real runs) or a permuted copy (for null runs). The guard `if ... and not null_control` correctly skips it for null runs. However, for **real runs**, `frame` is identical to `self._real_slice`, so the secondary outcomes are the *same for every strategy and every seed* on the same slice. This is documented ("IPS-unweighted in-sample treated/control-arm means") but it means `secondary_outcomes` carries no per-strategy information — it's a property of the dataset, not the optimizer output.

The docstring acknowledges this ("policy-conditioned secondary outcomes are deferred to a later sprint"), so this is by design. But you might want to consider whether storing redundant identical dicts on every result is worth the memory/JSON bloat versus computing it once on the scenario.

#### 4. The `run_strategy` random path doesn't set `secondary_outcomes` consistently (nit)

**`hillstrom.py:555-562`** (random strategy path) and **`hillstrom.py:564-578`** (engine path) both proceed to the same `_secondary_arm_aggregates` call at line 593. This is fine — both paths share the tail of the method. Just confirming there's no code path that bypasses it for real runs.

#### 5. Script `_load_raw` doesn't validate the loaded frame (low risk)

**`scripts/hillstrom_benchmark.py:217-220`**: `_load_raw` blindly returns the frame from `pd.read_csv` or `pd.read_parquet`. The validation happens downstream in `load_hillstrom_slice` which checks for required columns. This is fine — the error message from `load_hillstrom_slice` is clear. But if someone passes a completely wrong file, the error will come from a different function than where the file was loaded, which could be confusing. Minor concern.

#### 6. Script's `_print_summary` doesn't handle the case where all policy values are NaN (cosmetic)

**`scripts/hillstrom_benchmark.py:163-164`**: `pvals = [r.policy_value for r in group if math.isfinite(r.policy_value)]` — if all results in a group have NaN policy values, `pvals` is empty and `_fmt_mean_std` returns `"N/A"`. This is handled correctly. Good.

#### 7. `__init__.py` has a duplicate import block (style)

**`__init__.py:29-43`**: The Hillstrom imports are split into two `from causal_optimizer.benchmarks.hillstrom import (...)` blocks — one for most symbols and a separate one for renaming `VALID_STRATEGIES` to `HILLSTROM_VALID_STRATEGIES`. This is presumably to keep the rename separate, but it would be cleaner as a single import block:

```python
from causal_optimizer.benchmarks.hillstrom import (
    HILLSTROM_FROZEN_PARAMS,
    ...,
    VALID_STRATEGIES as HILLSTROM_VALID_STRATEGIES,
)
```

Ruff may or may not flag this; it's a minor style issue.

#### 8. Test `test_parquet_path` uses monkey-patching via attribute assignment instead of `monkeypatch` (test quality)

**`test_hillstrom_benchmark_script.py:255-268`**: The parquet test manually assigns `pd.read_parquet = fake_read_parquet` with a try/finally instead of using pytest's `monkeypatch` fixture. The other tests in this file use `monkeypatch` properly. This is inconsistent and slightly fragile — if the test crashes between assignment and the finally block in certain edge cases, the original function wouldn't be restored (though in practice `finally` handles this). Consider using `monkeypatch.setattr(pd, "read_parquet", fake_read_parquet)`.

#### 9. Missing `descriptor_names` in `ExperimentEngine` constructor (potential optimization miss)

**`hillstrom.py:565-572`**: The `ExperimentEngine` is constructed without `descriptor_names`. The adapter returns metrics like `total_cost` and `treated_fraction` that could serve as behavioral descriptors for MAP-Elites diversity in the exploitation phase. Existing benchmarks (`counterfactual_energy.py`, `interaction_policy.py`) also don't pass `descriptor_names`, so this is consistent with current patterns. But if the Hillstrom benchmark ever runs budgets > 50 (exploitation phase), MAP-Elites won't have descriptors to work with.

Not a bug for Sprint 31 budgets (20/40/80 with 80 pushing into exploitation), but worth noting for future work.

#### 10. `max_skips=0` not passed to engine (minor inconsistency with other benchmarks)

**`hillstrom.py:563-572`**: Other benchmark scenarios pass `max_skips=0` to disable the off-policy skip logic (since policy evaluation is cheap). The Hillstrom scenario doesn't. This means the off-policy predictor may skip some evaluations during the optimization phase, which could reduce the effective number of evaluations below `budget`. This is probably fine — it's a valid design choice — but it differs from the `counterfactual_energy`, `interaction_policy`, and `dose_response` patterns.

---

### Positive Observations

1. **Contract-driven design**: Every invariant from the Sprint 31 contract has a corresponding test. The propensity pin-downs, projected graph enumeration, frozen parameter assertions, and null-control permutation properties are all verified.

2. **Excellent test isolation**: The `_check_active_params_invariant` function and its tests (`TestActiveParamsInvariant`) are a great defensive guard against the frozen-param leak scenario. The `monkeypatch` test for missing adapter variables with proper `cache_clear()` is thorough.

3. **Clean separation of concerns**: The "loader + wrapped runner" pattern avoids modifying `MarketingLogAdapter` at all. `HillstromPolicyRunner` is a simple, testable adapter that injects frozen params without mutating the caller's dict.

4. **Null-control design**: Computing `null_baseline` once from the real frame and reusing it (since permutation preserves the mean) is correct and well-documented.

5. **Shared `VALID_STRATEGIES`**: Consolidating the strategy set as a module-level constant and importing it in both the benchmark module and the CLI script is the right pattern — it avoids the silent-divergence risk.

6. **Test coverage is very thorough**: 499 lines of unit tests + 132 lines of integration tests + 305 lines of script tests for 674 lines of implementation code. The defensive guard tests (missing columns, leaked frozen params, missing active keys) are particularly valuable.

7. **The fixture CSV is well-designed**: 300 rows with realistic Hillstrom-schema data, non-trivial distribution across all three segments, and enough non-zero spend/visit/conversion values to produce meaningful metrics.

---

### Summary

This is a solid, well-tested PR. The major design decisions (loader + wrapped runner, projected prior graph, frozen parameters, null-control baseline) are all sound and well-documented. The actionable items are:

- **Consider** adding `max_skips=0` to match other benchmark patterns (item 10)
- **Consider** using `monkeypatch` in the parquet test instead of manual try/finally (item 8)
- **Consider** merging the duplicate import block in `__init__.py` (item 7)

None of these are blocking. The code is ready for the gauntlet.

---

datablogin:
@greptileai

datablogin:
Re-reviewed current HEAD (`c6f5f61`). I found two actionable issues to address before I’d be comfortable approving this one.

1. `[P1]` The Hillstrom causal graph is wired to the wrong objective node, so the causal path is effectively losing its graph signal.

   In `causal_optimizer/benchmarks/hillstrom.py`, `hillstrom_projected_prior_graph()` points to `policy_value`, but `HillstromScenario.run_strategy()` builds `ExperimentEngine(..., objective_name="objective", minimize=True)` and reads `best_result("objective", minimize=True)`. I sanity-checked this locally: `hillstrom_projected_prior_graph().ancestors("objective")` and `.parents("objective")` are both empty, while the graph does have the expected ancestors/parents for `policy_value`.

   That matters because the causal optimizer internals key off `objective_name`: in `suggest.py`, causal focus / parent weighting use `ancestors(objective_name)` and `parents(objective_name)`, and in `off_policy.py` the observational path returns `None` when `objective_name` is not a node in the graph. So the current Hillstrom `causal` run is much closer to `surrogate_only` than intended.

   I’d fix this by making the graph and engine objective name consistent, the same way the other benchmark modules do it. Either:
   - retarget the Hillstrom graph to `objective`, or
   - run the engine directly on `policy_value` with consistent maximize/minimize semantics.

2. `[P2]` `secondary_outcomes` is still documented as policy-matched, but the implementation computes full-slice treated/control means.

   The dataclass comment on `HillstromBenchmarkResult.secondary_outcomes` says these are aggregates "taken over the logged observations matched by the best policy," but `run_strategy()` currently calls `_secondary_arm_aggregates(frame)` on the full slice, and `_secondary_arm_aggregates()` explicitly says it does *not* filter to the policy’s selected observations.

   That leaves the field contract inconsistent with the actual output. I’m okay with the simpler full-slice diagnostic if that’s the intended scope for Sprint 31, but then the field/docs/PR language should be softened to match. Otherwise the implementation needs to compute policy-matched secondary outcomes.

I did not duplicate Greptile’s smaller provenance/style notes here; the two items above are the ones I think matter most for correctness/contract fidelity.

datablogin:
Both items addressed in `a2e0301`. Good catch on the P1 — that was a real bug, not a docs issue, and the existing test suite did not surface it.

## [P1] — Causal graph now targets the engine's objective_name

Root cause confirmed: `HillstromScenario.run_strategy` constructed the engine with `objective_name="objective"` and the runner flipped sign to `-policy_value`, but `hillstrom_projected_prior_graph()` kept `policy_value` as the sink node. Your sanity check (`graph.ancestors("objective") == set()` and `graph.parents("objective") == set()`) is exactly what I see. That routed the causal path through `optimizer/suggest.py`'s `ancestors(objective_name)` and `parents(objective_name)` calls against an empty set, and through `predictor/off_policy.py`'s observational short-circuit when `objective_name` is not a node — the causal strategy was effectively running as surrogate_only.

**Fix:** dropped the sign-flipping alias in `HillstromPolicyRunner.run` and ran the engine directly with `objective_name="policy_value", minimize=False`. This matches `MarketingLogAdapter.get_minimize()` (which already returns `False`), keeps the projected graph as a strict subset of the full adapter graph (the existing `test_full_adapter_graph_has_all_seven_projected_edges` check still passes), and makes the engine's ancestor/parent lookups receive the intended 7-edge sub-DAG. The objective-name string is pinned as a module-level constant `_HILLSTROM_ENGINE_OBJECTIVE` so the graph sink node and the engine argument cannot drift again.

**Regression guard:** added `TestProjectedPriorGraph.test_graph_sink_node_is_the_engine_objective` that asserts (i) `_HILLSTROM_ENGINE_OBJECTIVE` is in `graph.nodes`, (ii) `graph.ancestors(...)` is non-empty, (iii) `graph.parents(...)` is non-empty, and (iv) at least one active search variable (`eligibility_threshold` / `regularization` / `treatment_budget_pct`) appears in the ancestor set. Any of those failing means the causal path has lost its graph signal.

**Verification under the fix:**
- `hillstrom_projected_prior_graph().ancestors("policy_value")` is now `{eligibility_threshold, regularization, treated_fraction, treatment_budget_pct}`, up from the empty set under the old wiring.
- `hillstrom_projected_prior_graph().parents("policy_value")` is `{regularization, treated_fraction}`.
- 2-seed CLI smoke run under `ax_botorch`: engine log now shows positive best-objective values (e.g. `best=4.4194`) instead of the old negated `best=-4.4194`, confirming the engine is maximizing `policy_value` directly.

## [P2] — `secondary_outcomes` field comment softened to match implementation

You are right that the field contract and implementation disagreed. The simpler full-slice diagnostic is the intended scope for Sprint 31 — policy-filtered aggregates need learned uplift/CATE scoring that is explicitly deferred in contract Section 4c. I took the softening path rather than adding policy filtering.

Field comment on `HillstromBenchmarkResult.secondary_outcomes` now reads:

> Secondary reported outcomes: **full-slice** treated/control-arm means of `visit` and `conversion` on the reshaped frame. These are *not* policy-conditioned — they are properties of the dataset, not of the optimizer output, and they are therefore identical across strategies and seeds on the same slice. Policy-filtered secondary outcomes (uplift- or CATE-conditioned subpopulations) are deferred to a later sprint per Sprint 31 contract Section 4c. See `_secondary_arm_aggregates` for the exact computation.

This cross-references the helper function whose docstring was already truthful after the earlier claudeloop rename.

## Test status

- **86** focused hillstrom tests pass (+1 new regression guard), **100%** line coverage on `causal_optimizer/benchmarks/hillstrom.py`.
- **1137** passed on the full fast suite (`-m "not slow"`), 0 failures.
- `ruff check` / `ruff format --check` / `mypy` all clean.
- CI currently running on `a2e0301`.

Head is now `a2e0301`. Ready for re-review.

datablogin:
Re-reviewed current HEAD (`a2e0301`). The objective wiring fix is good from my side, and the secondary-outcome wording now matches the implementation. I also sanity-checked the new graph coupling locally: `ancestors(policy_value)` and `parents(policy_value)` are both non-empty now, and `uv run pytest tests/unit/test_hillstrom_benchmark.py -q` passes.

One small defensive issue still remains:

1. `[P2]` `scripts/hillstrom_benchmark.py` still writes `hillstrom_provenance` without passing it through `_sanitize_for_json` before `json.dump(..., allow_nan=False)`.

   This is only reachable on a degenerate input, but it is real. A `primary` slice can be empty if the CSV contains only `Mens E-Mail` rows; I reproduced that locally and `HillstromScenario(..., slice_type=PRIMARY).null_baseline` becomes `nan`. That `nan` then flows straight into `null_baselines` at `scripts/hillstrom_benchmark.py:330-336`, so the artifact write can still fail at the final `json.dump(..., allow_nan=False)`.

   I’d fix this by sanitizing `hillstrom_provenance` (or at least the `null_baselines` leaf) before attaching it to `output_data["provenance"]`.

So: the two previous review items are fixed, but I would still patch this last JSON-serialization edge case before merge.

datablogin:
Fixed in `bd11809`. Wrapped the entire `hillstrom_provenance` dict in `_sanitize_for_json()` before attaching it to `output_data["provenance"]`, so any `nan` from an empty primary slice (or future degenerate input) is converted to `None` before `json.dump(..., allow_nan=False)`. Added a regression test that builds the exact provenance shape with a `nan` baseline and verifies both the sanitizer output and a round-trip through `json.dumps(allow_nan=False)`.

87 focused tests pass. CI running on `bd11809`. Ready for re-review.

datablogin:
Re-reviewed current HEAD (`bd11809`). The provenance sanitization fix is in place from my side.

What I checked:
1. `scripts/hillstrom_benchmark.py` now runs `hillstrom_provenance` through `_sanitize_for_json(...)` before attaching it to the artifact payload, so an empty-slice `null_baseline = nan` becomes `None` instead of tripping `json.dump(..., allow_nan=False)`.
2. the new CLI regression test for the `nan -> None` provenance case is present and I reran `uv run pytest tests/unit/test_hillstrom_benchmark_script.py -q` locally -> `35 passed`.

No blocking findings remain from my review. As of this pass, the remaining gate is CI: `build`, `lint`, and `typecheck` were green, while the matrix test jobs and `test-bayesian` were still running when I checked.

datablogin:
@greptileai

Issue #185, opened 2026-04-20
Sprint 35.A: implement the Open Bandit DomainAdapter and smoke test


datablogin:
## Context

Executes Issue A from Section 10 of the [Sprint 34 Open Bandit contract](../blob/main/thoughts/shared/docs/sprint-34-open-bandit-contract.md) and [Sprint 35 plan](../blob/main/thoughts/shared/plans/25-sprint-35-recommendation.md).

Sprint 34 contract is merged (PR #184). Sprint 35 starts implementation. This issue is the adapter-critical path for the first logged multi-action policy lane (ZOZOTOWN Men / uniform-random).

## Goal

Implement the Open Bandit `DomainAdapter` subclass required by the Sprint 34 contract, wire it to OBP as an optional extra, and complete the mandatory smoke test. Do **not** subclass `MarketingLogAdapter`.

## Tasks

1. Add an optional Open Bandit dependency extra to `pyproject.toml`.
2. Implement a new Open Bandit `DomainAdapter` subclass under `causal_optimizer/domain_adapters/` (e.g., `bandit_log.py`).
3. Load the Men/Random slice via OBP (`OpenBanditDataset`).
4. Implement the Section 4 search-space shape:
   - `tau` (softmax temperature, continuous)
   - `eps` (exploration epsilon, continuous)
   - 3-5 context-feature weights (continuous)
   - `position_handling_flag` (categorical: `marginalize` / `position_1_only`)
5. Complete the smoke test:
   - verify Men/Random row count is in the expected ~452,949 range (Saito et al. 2021 Table 1)
   - confirm whether `action_prob` is conditional (`P(item|position)`) or joint (`P(item,position)`)
   - document the chosen 3-5 context features in the adapter docstring
6. Fail fast with a clear error if OBP is missing.
7. Keep adapter surface narrow enough for Track B to consume without reopening architecture.

## File Ownership (disjoint with Track B)

Track A owns:
- `pyproject.toml`
- new adapter module under `causal_optimizer/domain_adapters/`
- `causal_optimizer/domain_adapters/__init__.py`
- adapter-focused tests under `tests/unit/` and `tests/integration/`
- any adapter-focused smoke-test helper

Track A must not edit `causal_optimizer/benchmarks/`.

## Out of scope

- benchmark report (Issue C)
- Women/All campaigns
- BTS as primary logger
- DRos, Switch-DR, MIPS, slate OPE
- action embeddings, per-item scoring, learned positional weighting

[...]

PR #188, opened 2026-04-20
Sprint 35.B: implement Open Bandit SNIPW/DM/DR stack and Section 7 gates

### LINKED ISSUE #186: Sprint 35.B: implement the Open Bandit SNIPW / DM / DR stack and Section 7 gates
## Context

Executes Issue B from Section 10 of the [Sprint 34 Open Bandit contract](../blob/main/thoughts/shared/docs/sprint-34-open-bandit-contract.md) and [Sprint 35 plan](../blob/main/thoughts/shared/plans/25-sprint-35-recommendation.md).

Sprint 34 contract is merged (PR #184). This issue is the evaluator-critical path for the first logged multi-action policy lane.

This issue runs in parallel with Issue A. Work against synthetic bandit-feedback dicts and stub policy outputs so you can make real progress before the adapter lands.

## Goal

Implement the Open Bandit OPE wrapper required by the Sprint 34 contract:
- SNIPW primary
- DM + DR secondary diagnostics
- Section 7 support gates

## Tasks

1. Implement the Open Bandit OPE wrapper under `causal_optimizer/benchmarks/` (e.g., `open_bandit.py`).
2. Implement estimators:
   - SNIPW (primary, gates the verdict)
   - DM (secondary diagnostic, expected to be biased)
   - DR (secondary diagnostic)
3. Implement the Section 7 gates:
   - 7a null control (position-stratified reward permutation, 5% relative band)
   - 7b ESS floor (median ESS >= `max(1000, n_rows/100)`)
   - 7c zero-support fraction (<= 10%)
   - 7d propensity sanity (10% relative band against schema-appropriate target)
   - 7e DR/SNIPW cross-check (<= 25% relative divergence per seed)
4. Report the Section 4d diagnostics: `policy_value`, `ess`, `weight_cv`, `max_weight`, `zero_support_fraction`, `n_effective_actions`.
5. Start against synthetic bandit-feedback dicts and stub policy outputs. Keep the interface adapter-agnostic so final wiring is a small follow-up, not a rewrite.
6. Ship focused unit and integration coverage.

## File Ownership (disjoint with Track A)

Track B owns:
- new Open Bandit OPE / benchmark wrapper under `causal_optimizer/benchmarks/`
- OPE- and gate-focused tests under `tests/unit/` and `tests/integration/`
- synthetic bandit-feedback fixtures and helpers for gate testing

Track B must not edit `pyproject.toml` or the adapter package.

## Out of scope

- `pyproject.toml` edits (Track A)
- adapter implementation (Track A)
- the full Men/Random benchmark report (Issue C)
- DRos, slate OPE, Women, All, BTS-primary

## Workflow

`/tdd` -> implement -> `/polish` -> `gh pr create` -> `/gauntlet` -> report PR URL. Do not merge. Leave open for human review.

## Deliverables

1. PR URL
2. which gates are implemented
3. synthetic fixtures / helpers added
4. exact commands run
5. gauntlet status

datablogin:
## Summary

Implements the Sprint 34 Open Bandit contract (Sections 4d, 5, 6, 7) for Issue #186. This is Track B of the Sprint 35 parallel build: the OPE evaluator and support gates, built adapter-agnostic against synthetic bandit-feedback dicts so Track A's adapter (Issue #185) can drop in without rework.

- **Estimators (Section 5a):** SNIPW primary verdict estimator, DM and DR secondary diagnostics (in-module Dudík et al. 2011 DR formula; OBP DR hook documented for when the optional extra lands).
- **Propensity floor (Section 5c):** fixed `min_propensity_clip = 1 / (2 * n_items * n_positions)` per contract. Frozen (not tuned); clipped row count reported.
- **Section 4d diagnostics:** `policy_value`, `ess`, `weight_cv`, `max_weight`, `zero_support_fraction`, `n_effective_actions`, `n_clipped_rows`, `estimator` (provenance).
- **Section 7 gates (all five):**
  - **7a** null control — position-stratified reward permutation under a fixed seed; fails if any strategy exceeds `1.05 * mu_null`.
  - **7b** ESS floor — median per-seed ESS must be `>= max(1000, n_rows / 100)`.
  - **7c** zero-support fraction — best-of-seed must be `<= 10%`.
  - **7d** propensity sanity — empirical `action_prob` mean within `10%` relative of the schema-appropriate target (`"conditional"` vs `"joint"`; schema is an explicit parameter, not hardcoded — Track A's smoke test will confirm which the OBD loader uses).
  - **7e** DR/SNIPW cross-check — any seed with `>25%` relative divergence fails the gate and flags the seed index.
- **`run_section_7_gates`** bundler returns a `GateReport` dataclass with an `all_passed()` helper and a provenance dict (including an OBP version placeholder).

Public entry points accept a `bandit_feedback` dict (OBP schema: `n_rounds`, `n_actions`, `action`, `reward`, `pscore`, `position`) plus a `[n_rounds, n_actions]` action distribution. Synthetic fixtures (`generate_synthetic_bandit_feedback`, `uniform_policy`, `peaked_policy`, `degenerate_policy`, `permute_rewards_stratified`) let Track A and Track C test against the same shape without OBP installed.

## Gates implemented and test status

| Gate | Implemented | Pass-case test | Fail-case test |
| --- | --- | --- | --- |
| 7a null control (position-stratified, 5% relative band) | yes | `test_null_control_gate_passes_on_permuted_data` | `test_null_control_gate_fails_when_strategy_exceeds_band` |
| 7b ESS floor (`max(1000, n_rows/100)`) | yes | `test_ess_floor_formula`, `test_ess_floor_lower_bound_1000` | `test_ess_gate_fails_when_median_below_floor` |
| 7c zero-support fraction (<= 10%) | yes | `test_zero_support_gate_passes_below_threshold` | `test_zero_support_gate_fails_when_best_exceeds_10pct` |
| 7d propensity sanity (10% relative) | yes | `test_joint_schema_target`, `test_conditional_schema_target`, `test_within_band_passes` | `test_outside_10pct_relative_band_fails`, `test_propensity_sanity_fails_outside_band` |
| 7e DR/SNIPW cross-check (<= 25% relative) | yes | `test_matching_values_pass` | `test_large_divergence_fails`, `test_single_seed_over_threshold_fails`, `test_zero_baseline_handled_gracefully` |

## Files added

- `causal_optimizer/benchmarks/open_bandit.py` — OPE wrapper, estimators, Section 7 gates, and synthetic fixtures (259 lines + docstrings).
- `tests/unit/test_open_bandit_ope.py` — 62 unit tests (estimators, diagnostics, gates, fixtures, input validation).
- `tests/integration/test_open_bandit_gates.py` — 10 integration tests (end-to-end passing run, per-gate failure cases, SNIPW sanity).

Also re-exports the new symbols from `causal_optimizer.benchmarks.__init__` so downstream Issue A and Issue C imports stay tidy.

## Notes for Track A (Issue #185) integration

The module is adapter-agnostic. Track A's adapter should build a `bandit_feedback` dict of the same shape (keys: `n_rounds`, `n_actions`, `action`, `reward`, `pscore`, `position`) from OBP's loader and feed it straight into `evaluate_open_bandit_policy`. No shared files were touched on the Track A side; final wiring is a small follow-up once the schema smoke test runs.

The `OBP_VERSION_PLACEHOLDER` constant in the provenance dict is the hook where Track A plugs in `obp.__version__`.

## Test plan

- [x] 72 new tests (62 unit + 10 integration) pass locally
- [x] Full fast suite: 1240 passed, 24 skipped (no regressions)
- [x] `uv run ruff check .` — clean
- [x] `uv run ruff format .` — clean
- [x] `uv run mypy causal_optimizer/benchmarks/open_bandit.py` — strict, clean
- [x] Patch coverage on `open_bandit.py`: 100%
- [ ] CI green
- [ ] claude-review.sh + greploop + human approval before merge

Closes #186

<h3>Greptile Summary</h3>

Implements the Sprint 34 Open Bandit OPE stack (SNIPW/DM/DR estimators, Section 4d diagnostics) and all five Section 7 quality gates as first-class dataclasses, backed by 72 new tests (100% patch coverage). All three issues from the prior review round — `peaked_policy` lower-bound guard, `reward_mean_by_action` range validation, and the `bf` parameter name in `run_section_7_gates` — have been addressed.

Safe to merge; only P2 findings remain, both covering edge cases unlikely in practice.

All P0/P1 issues from the prior review are resolved. The two new findings are P2: one covers a latent inconsistency when a caller deliberately passes n_actions values that disagree with bandit_feedback["n_actions"], and one covers vacuous-true behavior on an empty strategy_policies dict. Neither affects any current call site. Test coverage is comprehensive with hand-computed reference checks.

No files require special attention; both P2 comments are in open_bandit.py.

| Filename | Overview |
|----------|----------|
| causal_optimizer/benchmarks/open_bandit.py | New 945-line module implementing SNIPW/DM/DR estimators and all five Section 7 gates; logic is sound with two minor inconsistencies flagged (n_actions mismatch across run_section_7_gates, vacuous pass on empty strategy_policies). |
| tests/unit/test_open_bandit_ope.py | 62 unit tests with hand-computed reference checks, boundary cases, and all ValueError/KeyError paths; coverage appears comprehensive. |
| tests/integration/test_open_bandit_gates.py | 10 end-to-end integration tests covering all-pass scenario, each gate's failure path, and GateReport structure; no issues found. |
| causal_optimizer/benchmarks/__init__.py | Re-exports all new open_bandit symbols; alphabetical ordering also fixes a pre-existing load_criteo_subsample / load_energy_frame swap. No issues. |
| .gitignore | Adds .coverage to the ignore list; no issues. |

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    BF["bandit_feedback dict\n(n_rounds, n_actions, action,\nreward, pscore, position)"] --> EVL["evaluate_open_bandit_policy\n(Section 4d diagnostics)"]
    BF --> G7["run_section_7_gates"]
    EVL --> SNIPW["compute_snipw\n(primary verdict)"]
    EVL --> DM["compute_dm\n(diagnostic)"]
    EVL --> DR["compute_dr\n(diagnostic)"]
    G7 --> GA["7a null_control_gate\nSNIPW ≤ 1.05 × μ_null"]
    G7 --> GB["7b ess_gate\nmedian(ESS) ≥ max(1000, n/100)"]
    G7 --> GC["7c zero_support_gate\nbest-of-seed ≤ 10%"]
    G7 --> GD["7d propensity_sanity_gate\n|empirical − target| / target ≤ 10%"]
    G7 --> GE["7e snipw_dr_cross_check_gate\n|SNIPW − DR| / max ≤ 25%"]
    GA --> RPT["GateReport\n.all_passed() · .provenance"]
    GB --> RPT
    GC --> RPT
    GD --> RPT
    GE --> RPT
    CLIP["compute_min_propensity_clip\n1/(2·n_actions·n_positions)"] --> EVL
    CLIP --> GA
```

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%202%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%202%0Acausal_optimizer%2Fbenchmarks%2Fopen_bandit.py%3A882-930%0A**%60n_actions%60%20%2F%20%60n_positions%60%20inconsistency%20between%20gates**%0A%0A%60run_section_7_gates%60%20takes%20explicit%20%60n_actions%60%20and%20%60n_positions%60%20parameters%20and%20forwards%20them%20to%20%60propensity_sanity_gate%60%2C%20but%20%60null_control_gate%60%20ignores%20those%20values%20entirely%20%E2%80%94%20it%20reads%20%60n_actions%60%20directly%20from%20%60bandit_feedback%5B%22n_actions%22%5D%60%20and%20infers%20%60n_positions%60%20from%20%60position.max%28%29%20%2B%201%60.%20If%20a%20caller%20passes%20a%20mismatched%20%60n_actions%60%20%28e.g.%20passes%20%60n_actions%3D34%60%20for%20a%20%60bandit_feedback%60%20whose%20%60%22n_actions%22%60%20key%20is%20%6010%60%29%2C%20the%20propensity%20sanity%20target%20%28%601%2F%28n_actions%20*%20n_positions%29%60%29%20and%20the%20clip%20floor%20used%20internally%20by%20the%20null-control%20gate%20will%20be%20computed%20against%20two%20different%20action-space%20sizes%2C%20silently%20producing%20inconsistent%20gate%20results.%0A%0A%23%23%23%20Issue%202%20of%202%0Acausal_optimizer%2Fbenchmarks%2Fopen_bandit.py%3A703-720%0A**Vacuous%20pass%20when%20%60strategy_policies%60%20is%20empty**%0A%0A%60all%28v%20%3C%3D%20threshold%20for%20v%20in%20%7B%7D.values%28%29%29%60%20evaluates%20to%20%60True%60%20by%20Python's%20vacuous-truth%20rule.%20Calling%20%60null_control_gate%28...%2C%20strategy_policies%3D%7B%7D%2C%20...%29%60%20silently%20returns%20%60passed%3DTrue%60%20even%20though%20no%20policy%20was%20actually%20checked.%20A%20guard%20such%20as%20%60if%20not%20strategy_policies%3A%20raise%20ValueError%28...%29%60%20would%20make%20the%20caller%20error%20explicit%20rather%20than%20silently%20producing%20a%20misleading%20%22passed%22%20result.%0A%0A&repo=datablogin%2Fcausal-optimizer&pr=188&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=2">Fix All in Claude Code</picture></a>

`````markdown
Path: causal_optimizer/benchmarks/open_bandit.py
Line: 882-930

Comment:
**`n_actions` / `n_positions` inconsistency between gates**

`run_section_7_gates` takes explicit `n_actions` and `n_positions` parameters and forwards them to `propensity_sanity_gate`, but `null_control_gate` ignores those values entirely — it reads `n_actions` directly from `bandit_feedback["n_actions"]` and infers `n_positions` from `position.max() + 1`. If a caller passes a mismatched `n_actions` (e.g. passes `n_actions=34` for a `bandit_feedback` whose `"n_actions"` key is `10`), the propensity sanity target (`1/(n_actions * n_positions)`) and the clip floor used internally by the null-control gate will be computed against two different action-space sizes, silently producing inconsistent gate results.

---

Path: causal_optimizer/benchmarks/open_bandit.py
Line: 703-720

Comment:
**Vacuous pass when `strategy_policies` is empty**

`all(v <= threshold for v in {}.values())` evaluates to `True` by Python's vacuous-truth rule. Calling `null_control_gate(..., strategy_policies={}, ...)` silently returns `passed=True` even though no policy was actually checked. A guard such as `if not strategy_policies: raise ValueError(...)` would make the caller error explicit rather than silently producing a misleading "passed" result.

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["address claude review feedback (claudelo..."](https://github.com/datablogin/causal-optimizer/commit/0b5fac49fa54e7fa38edaf20327ba6a31b453040) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28929569)</sub>

### CODE REVIEW

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the code. Here's my review:

---

## PR Review: Sprint 35.B — Open Bandit SNIPW/DM/DR Stack and Section 7 Gates

### Overall Assessment

This is a well-structured, carefully implemented module. The code follows existing benchmark patterns (Hillstrom, Criteo), the estimator math is correct, the gate logic is clean, and test coverage is comprehensive. There are a few issues worth addressing, ranging from a potential bug to minor style nits.

---

### Bugs / Correctness Issues

**1. `null_control_gate` uses wrong data for `mu_null` computation (medium severity)**

`open_bandit.py:646` — The null control gate computes `mu_null` as the mean of the *permuted* rewards, but then evaluates SNIPW on the permuted data too. The gate's contract (Section 7a) says strategies should not be able to extract signal from permuted (null) data. The comparison `v <= band_multiplier * mu_null` checks that SNIPW on permuted data doesn't significantly exceed the permuted reward mean.

This is actually correct for a uniform logger + uniform eval policy (SNIPW collapses to the sample mean). However, for non-uniform eval policies, the SNIPW estimate on permuted data may legitimately differ from `mu_null` (the raw reward mean) even under the null, because SNIPW reweights. The band should arguably be around the *SNIPW value under the null policy* (e.g., SNIPW of a uniform policy on permuted data), not the raw reward mean. As-is, the gate may produce false positives for peaked policies on permuted data, since SNIPW reweighting can push the estimate above 1.05x the raw mean even without signal.

For Track B with synthetic uniform loggers this won't manifest, but worth noting for Track A with real OBD data.

**2. `n_effective_actions` off-by-one edge case (low severity)**

`open_bandit.py:468` — `np.searchsorted(cumsum, threshold, side="left")` returns the index of the first value `>= threshold`. For a uniform policy with 5 actions, cumsum is `[0.2, 0.4, 0.6, 0.8, 1.0]`. Searching for 0.95 returns index 4 (since 1.0 >= 0.95), so `n_effective_actions = 5`. This is correct.

But if the threshold is *exactly* hit (e.g., threshold=0.8, cumsum=[0.2, 0.4, 0.6, 0.8, 1.0]), `searchsorted(side="left")` returns index 3, giving `n_effective_actions = 4`. If the intent is "smallest k whose cumulative mass *exceeds*" (strict), this should use `side="right"` instead. The current behavior says "smallest k whose cumulative mass *reaches or exceeds*". The docstring says "exceeds" — consider aligning.

**3. ESS floor uses integer division (low severity)**

`open_bandit.py:696` — `n_rows // 100` truncates. The docstring says `n_rows / 100`. For `n_rows=199`, `199 // 100 = 1` but `199 / 100 = 1.99`. Since the floor has a lower bound of 1000 this won't matter in practice, but the implementation doesn't match the docstring. Consider using `n_rows / 100` with `int(max(1000, n_rows / 100))` or updating the docstring.

---

### Design / API Issues

**4. `null_control_gate` falls back to `position=zeros` but `permute_rewards_stratified` requires `position`**

`open_bandit.py:648` — When computing `n_positions`, the function does `bandit_feedback.get("position", np.zeros(...))`, suggesting position is optional. But the very first call (`permute_rewards_stratified` at line 645) will `KeyError` if position is missing. This is an unreachable fallback — the function will always fail earlier. Either make position truly required (remove the `.get` fallback) or move the permutation call after the position check.

**5. `strategy_value_overrides` is test-only machinery exposed in the public API**

`open_bandit.py:639-643` — The `strategy_value_overrides` parameter on `null_control_gate` exists solely for test convenience. This works, but an alternative pattern would be to test the gate by constructing deterministic synthetic data that naturally triggers failure, keeping the public API clean. Not a blocker, but worth considering for Track A.

**6. `PropensitySanityResult.schema` typed as `str` instead of `PropensitySchema`**

`open_bandit.py:577` — The field is `schema: str` but only `"conditional"` or `"joint"` are valid. Using `PropensitySchema` (the `Literal` type) would be more precise and consistent with the function signature that creates it.

**7. `__init__.py` `__all__` sorting is inconsistent**

The `__all__` list mixes constants (uppercase), classes (PascalCase), and functions (snake_case) into a single sorted block. The existing pattern in this repo's `__init__.py` groups constants first, then classes, then functions — but the new entries break this pattern in a few spots (e.g., `"DEFAULT_PROPENSITY_SCHEMA"` appears between `"CriteoScenario"` and `"HILLSTROM_FROZEN_PARAMS"`). The list does appear to be mostly alphabetically sorted, which is fine, but the interleaving of constants within the class block at lines 156-163 looks off.

---

### Numerical Stability

**8. `denom == 0.0` exact float comparison in SNIPW**

`open_bandit.py:333` — The `if denom == 0.0` check is safe here because when all `pi_e` values are exactly 0.0 (degenerate policy on non-logged actions), the weights are exactly 0.0 and their sum is exactly 0.0. There's no floating-point accumulation that could produce a near-zero non-zero sum. This is correct.

**9. ESS denominator guard**

`open_bandit.py:454` — `max(float((weights**2).sum()), 1e-300)` is a reasonable guard against division by zero. The `1e-300` floor is well below any realistic weight magnitude. Good.

---

### Test Coverage

Test coverage is strong. Every public function has at least one passing and one failing test. Specific observations:

**10. Integration test `TestEndToEndPassingRun` uses DR == SNIPW (no real DR test)**

`test_open_bandit_gates.py:65` — The comment says "DR == SNIPW (dummy reward model = 0)", but `dr_per_seed` is set to the SNIPW value, not an actual DR computation. This means the cross-check gate is trivially passing (0% divergence). Consider computing actual DR values with a non-trivial `reward_hat` to exercise the cross-check gate meaningfully.

**11. No test for `GateReport.all_passed() == False`**

There's a test for `all_passed() == True` but no explicit test for `all_passed() == False` when at least one gate fails. The individual gate failure tests cover the semantics, but an integration test showing `all_passed()` returns `False` would be a good safety net.

**12. `compute_dr` has no dedicated fuzz/property test**

The DR estimator is the most complex formula. The unit test (`test_dr_falls_back_to_dm_when_pscore_is_exact`) only tests the degenerate case where residuals are zero. A test with non-zero residuals confirming DR sits between DM and SNIPW (as expected by theory) would strengthen confidence.

---

### Style / Minor

**13. Unused import in test file**

`test_open_bandit_ope.py:40` — `PropensitySchema` is imported but only used in a smoke test (`test_propensity_schema_literal_type`). This is fine for a smoke test; just noting it.

**14. `uniform_policy` takes positional args while `peaked_policy` / `degenerate_policy` use keyword-only**

`open_bandit.py:215` vs `228` — `uniform_policy(n_rounds, n_actions)` allows positional args while `peaked_policy(*, n_rounds, ...)` and `degenerate_policy(*, n_rounds, ...)` force keyword-only. This inconsistency is minor but could confuse callers. Consider making `uniform_policy` keyword-only too for API consistency.

---

### Summary

| Category | Count |
|----------|-------|
| Potential bugs | 2 (null_control mu_null semantics, n_effective_actions off-by-one) |
| Design issues | 3 (unreachable fallback, test-only API param, str vs Literal type) |
| Test gaps | 3 (trivial DR cross-check, no all_passed False test, DR fuzz) |
| Style nits | 2 (positional vs kw-only, __all__ ordering) |

The code is well-written and production-ready for Track B. The mu_null semantics issue (#1) is the most important item to think about before Track A lands real OBD data. The rest are improvements that can land incrementally.

---

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the code. Here's my review:

---

## PR Review: Sprint 35.B — Open Bandit SNIPW/DM/DR Stack and Section 7 Gates

Overall this is a well-structured, well-documented module. The estimator math is correct, the gate logic is clean, and test coverage is thorough. I have a mix of substantive issues and smaller suggestions.

---

### Substantive Issues

**1. `null_control_gate` computes `mu_null` from permuted rewards, but evaluates strategies against permuted data too — the gate semantics are subtle and may be wrong in one edge case (line 671)**

When `mu_null == 0.0`, the ratio is set to `inf` but `threshold = 1.05 * 0.0 = 0.0`. Any strategy with a positive SNIPW value will fail the gate. This is arguably correct (if the null baseline is zero, any positive estimate is suspicious), but worth documenting explicitly. If all rewards are zero, every SNIPW estimate is also zero and the gate passes — fine. But if permutation produces `mu_null = 0.0` while SNIPW is very small but positive (floating point), you get a false failure. Consider adding a guard or documenting this edge case.

**2. `evaluate_open_bandit_policy` returns `dict[str, Any]` — consider a typed return (lines 399-504)**

The rest of the benchmarks module uses typed dataclasses for results (`CriteoBenchmarkResult`, `HillstromBenchmarkResult`, etc.). This function returns an untyped dict. While it's consistent with the `bandit_feedback` input convention, a `PolicyEvaluationResult` dataclass would be more type-safe and make downstream consumers' lives easier. This is a design choice rather than a bug, but worth considering as the module matures.

**3. `compute_dr` does not validate shape consistency between `action_dist`, `reward_hat`, and `bandit_feedback` (lines 357-393)**

`compute_dm` validates that `action_dist.shape == reward_hat.shape`, but `compute_dr` does not check that `reward_hat` matches `action_dist` in shape, nor that `bandit_feedback["action"]` length matches the row count. An out-of-bounds index on `reward_hat[np.arange(len(action)), action]` would produce a cryptic numpy IndexError rather than a helpful message.

**4. `n_positions` inference in `null_control_gate` assumes 0-indexed contiguous positions (line 654)**

```python
n_positions = int(position.max() + 1) if position.size > 0 else 1
```

If positions are not 0-indexed or have gaps (e.g., `[0, 2, 5]`), this overestimates `n_positions`, which in turn computes a too-small propensity clip floor. This is fine for synthetic data but could bite when real OBD data lands via Issue A. Worth a comment noting the assumption.

**5. `strategy_value_overrides` is a test-only hook in production API (lines 621-623)**

Exposing a test-only parameter in the public API surface is a code smell. Consider either:
- Prefixing with `_` in the docstring to signal "internal"
- Moving the override logic to a test helper that patches `compute_snipw`
- At minimum, the current docstring adequately warns production callers not to use it, which is acceptable for now.

---

### Numerical / Statistical Correctness

**6. SNIPW zero-denominator guard uses exact float comparison (line 334)**

```python
if denom == 0.0:
```

This is correct here because the only way `denom` is exactly zero is if every `pi_e` weight is exactly zero (the policy places zero mass on every logged action). Floating-point weights from division cannot accidentally land on exact zero. Good.

**7. ESS formula has a correct `1e-300` guard (line 455)**

```python
ess = float((weights.sum() ** 2) / max(float((weights**2).sum()), 1e-300))
```

The `1e-300` floor prevents division by zero when all weights are near-zero but not exactly zero. This is correct. The ESS formula `(sum w)^2 / sum(w^2)` is standard (Kish's effective sample size).

**8. DR formula matches Dudík et al. 2011** — I verified the math against the paper. The implementation is correct.

---

### Code Quality

**9. `__init__.py` `__all__` list is unsorted within sections (lines 153-238)**

The `__all__` list mixes alphabetical ordering with insertions. The diff shows the list is being maintained manually. `"DEFAULT_PROPENSITY_SCHEMA"` appears between `"CriteoScenario"` and `"HILLSTROM_FROZEN_PARAMS"` — this is correct alphabetically for constants (uppercase first), but the overall list has uppercase constants mixed with capitalized classes and lowercase functions in a single flat sort. This is fine and matches the existing pattern, just noting it's getting long.

**10. Module-level docstring is excellent** — the public API listing, the adapter-agnostic interface note, and the section cross-references are all helpful. Well done.

**11. `peaked_policy` normalization guard is good defensive code (lines 266-268)**

The explicit normalization after setting peak mass handles floating-point precision well.

---

### Test Coverage

**12. Tests are comprehensive.** Coverage includes:
- Hand-computed estimator values (SNIPW, DR)
- Boundary conditions (zero weights, degenerate policies, zero `mu_null`)
- Input validation for every public function
- Determinism under fixed seeds
- Immutability of input dicts
- Both passing and failing scenarios for every gate
- Integration test exercising the full `run_section_7_gates` pipeline

**13. Missing test: `compute_dr` with mismatched `reward_hat` shape**

There's a test for `compute_dm` shape mismatch but none for `compute_dr`. Since `compute_dr` doesn't validate this (see issue #3), adding a test would expose the gap.

**14. Integration test `test_all_gates_pass_on_uniform_logger_uniform_policy` uses the same policy for all 3 seeds (lines 58-64)**

The loop `for _seed in range(3)` creates the same uniform policy each iteration, so `per_seed_ess`, `snipw_per_seed`, etc. have identical values across seeds. This doesn't test the seed-varying behavior that the gate is designed for. It still validates the pipeline, but a more realistic test would vary the policy or data across seeds.

---

### Minor / Style

**15. `PropensitySchema` is a type alias, not exported cleanly for mypy**

```python
PropensitySchema = Literal["conditional", "joint"]
```

This works at runtime and with mypy, but `TypeAlias` (from `typing` or `typing_extensions`) would make intent clearer:

```python
PropensitySchema: TypeAlias = Literal["conditional", "joint"]
```

This is a nit — the current form works fine with Python 3.10+.

**16. `weight_cv` uses `ddof=0` (line 459)**

Population std with `ddof=0` is the right choice for a diagnostic on the full sample of weights (not estimating a population parameter from a sample). Correct.

**17. `n_effective_actions` edge case: when all mass is on one action and `threshold=1.0` (line 472)**

`min(idx + 1, n_actions)` correctly caps at `n_actions`. The `searchsorted` with `side="right"` correctly handles the boundary. Good.

---

### Summary

| Category | Rating |
|----------|--------|
| Statistical correctness | Strong — estimators match published formulas |
| Numerical stability | Good — clip floors, zero guards, normalization |
| API design | Good — keyword-only args, clear naming |
| Type safety | Adequate — `dict[str, Any]` for bandit feedback is pragmatic but untyped |
| Test coverage | Excellent — ~60 tests covering happy path, edge cases, validation |
| Code organization | Clean — logical sections, good separation of concerns |

The main actionable items are: (1) add shape validation to `compute_dr`, (2) document the `mu_null == 0` edge case in `null_control_gate`, and (3) consider a typed return for `evaluate_open_bandit_policy` as the module matures. Everything else is minor or a design preference.

---

datablogin:
# Claude Code Review

## Review Feedback

I have thoroughly reviewed the code. Here is my detailed review:

---

## PR Review: Sprint 35.B — Open Bandit SNIPW/DM/DR Stack and Section 7 Gates

### Overall Assessment

This is a well-structured, well-documented module. The statistical implementations are correct, the gate logic follows the contract faithfully, and the test coverage is thorough. There are several issues worth addressing before merge, ranging from a potential correctness bug to design improvements.

---

### Bugs / Correctness Issues

**1. Null control gate uses wrong data for SNIPW evaluation (potential bug)**
`open_bandit.py:683-684` — The null control gate evaluates strategies against the *permuted* feedback dict. This is intentional for the null-control check — but the docstring at line 638 says it "fails the gate if any strategy's value exceeds `band_multiplier * mu_null`." The issue is subtler: `mu_null` is the mean of the *permuted* rewards (line 661), and SNIPW is also computed on the permuted data. Under a uniform eval policy, SNIPW on permuted data equals `mean(permuted_reward)` = `mu_null`, so the gate trivially passes with ratio ~1.0. This means **the gate can only detect strategies whose SNIPW weights systematically amplify certain rewards** — but on *permuted* data where the action-reward link is destroyed, this is a very weak signal. Is this the intended design? The contract says "5% relative band" but the gate as implemented will almost never fire on well-behaved SNIPW weights. Consider whether the gate should evaluate strategies against the *original* (unpermuted) data and compare against `mu_null` from the permuted data.

**2. `n_effective_actions` off-by-one edge case**
`open_bandit.py:481` — With `side="right"`, when `cumsum[-1]` equals `effective_action_mass_threshold` exactly (floating point permitting), `searchsorted` returns the index *after* that entry, so `idx + 1` could exceed `n_actions`. The `min(idx + 1, n_actions)` guard catches this, but the semantic issue is that a policy whose total mass equals exactly the threshold is considered to *not* exceed it, requiring all actions. The docstring says "exceeds" which is correct for `side="right"`, but this edge case deserves a unit test to pin the behavior.

**3. `null_control_gate` — `mu_null == 0` makes the gate always pass**
`open_bandit.py:693-694` — When all permuted rewards are zero, `threshold = band_multiplier * 0.0 = 0.0`. Since SNIPW on all-zero rewards also returns 0.0, the gate passes. But `per_strategy_ratios` reports `inf` (line 691). This is internally consistent but surprising: the ratio says "infinity" yet the gate says "passed." Consider documenting this edge case or adding a warning.

---

### Numerical Stability

**4. ESS denominator guard uses `1e-300` rather than a relative epsilon**
`open_bandit.py:465` — `max(float((weights**2).sum()), 1e-300)` prevents division by zero, but `1e-300` is so small that it provides essentially no protection against catastrophic cancellation when `weights.sum()` is also extremely small. This is fine in practice since the `if weights.sum() > 0.0` branch wouldn't be entered with all-zero weights, but the guard is misleading — consider using `np.finfo(float).tiny` or removing the guard entirely since the `if` branch already handles the zero case.

**5. `weight_cv` uses population `std(ddof=0)`, not sample `std(ddof=1)`**
`open_bandit.py:469` — This is a deliberate choice (population CV over the full set of weights), which is fine, but it differs from the statistical convention of using `ddof=1` for sample CV. Worth documenting the choice explicitly, especially since downstream gates may compare CV values across different `n_rounds`.

---

### Type Safety / API Design

**6. `evaluate_open_bandit_policy` returns `dict[str, Any]` — consider a typed return**
The diagnostic dict has a well-defined schema (8 keys, fixed types). A `TypedDict` or a dataclass would make downstream usage safer and enable autocompletion. The gate result dataclasses are already typed — the evaluation output should be too.

```python
class PolicyEvaluationResult(TypedDict):
    policy_value: float
    ess: float
    weight_cv: float
    max_weight: float
    zero_support_fraction: float
    n_effective_actions: int
    n_clipped_rows: int
    estimator: str
```

**7. `estimator` parameter should use a `Literal` type**
`open_bandit.py:416` — `estimator: str = "snipw"` accepts arbitrary strings but only recognizes `"snipw"`, `"dm"`, `"dr"`. Use `Literal["snipw", "dm", "dr"]` for static checking.

**8. `bandit_feedback` dicts are untyped throughout**
Every function accepts `dict[str, Any]` for `bandit_feedback`. A `TypedDict` (or at minimum a module-level docstring defining the schema) would catch key typos at type-check time.

---

### Design / Architecture

**9. `strategy_value_overrides` is a test-only escape hatch in production code**
`open_bandit.py:633` — This parameter exists solely for tests to inject fake values. It's documented as "production callers should not supply this," but it's part of the public API surface and will appear in IDE autocomplete. Consider moving the override logic to the test itself (e.g., monkeypatching `compute_snipw`) or making the parameter private/underscore-prefixed: `_strategy_value_overrides`.

**10. `__init__.py` re-exports everything — very large public surface**
The `__init__.py` re-exports 28 new symbols. For a benchmarks sub-package, this may be intentional, but `PropensitySchema` (a `Literal` type alias) and `OBP_VERSION_PLACEHOLDER` (a placeholder string) feel like implementation details rather than user-facing API. Consider whether all 28 need to be in `__all__`.

**11. `run_section_7_gates` takes `n_actions` and `n_positions` redundantly**
These values are already present in `bandit_feedback["n_actions"]` and derivable from `bandit_feedback["position"]`. The caller must pass them separately, creating a consistency risk. Consider extracting them from `bandit_feedback` inside the function (as `null_control_gate` already does for `n_positions`).

---

### Test Coverage

**12. No test for `peaked_policy` with `n_actions=1`**
`open_bandit.py:263` — `max(n_actions - 1, 1)` handles the single-action edge case, but there's no test for it. A peaked policy with 1 action is degenerate and should be equivalent to degenerate_policy.

**13. No test for `compute_snipw` with non-uniform propensity scores**
All SNIPW tests use either uniform pscore (`0.5`, `1/n`) or inject pathological values for clipping tests. There's no test with heterogeneous propensities that would exercise the weight computation more realistically.

**14. Integration test `TestEndToEndPassingRun` uses identical seeds**
`test_open_bandit_gates.py:58` — The loop `for _seed in range(3)` creates 3 identical uniform policies and evaluates them on the same feedback. The `per_seed_ess` and `per_seed_zero_support` lists will contain 3 identical values. This tests the plumbing but not the statistical behavior across seeds. Consider varying the seed in `generate_synthetic_bandit_feedback` or the policy across iterations.

**15. Missing negative test for `propensity_sanity_gate` with `n_actions=0` or `n_positions=0`**
The gate function doesn't validate these inputs — it would silently produce `ZeroDivisionError` from `1.0 / (n_actions * n_positions)`. Either add validation or document that callers must ensure positive values.

---

### Performance

**16. `permute_rewards_stratified` uses a Python loop over positions**
`open_bandit.py:546` — The `for pos_value in np.unique(position)` loop is O(n_positions) Python iterations. For OBD with 3 positions this is negligible, but if this is reused on datasets with many positions, it could be slow. The current implementation is clear and correct; just noting the scaling.

**17. Weight computation is duplicated in `evaluate_open_bandit_policy` and `compute_snipw`**
`open_bandit.py:459-461` computes `weights = pi_e_logged / clipped_pscore`, then line 486 calls `compute_snipw` which recomputes the same weights internally. For the SNIPW path, this doubles the work. Consider passing pre-computed weights or extracting a shared helper.

---

### Minor / Style

**18. `__all__` sort order in `__init__.py` is inconsistent**
The `__all__` list mixes constants, classes, and functions without a clear grouping convention. The existing benchmarks seem to follow alphabetical order, which the new entries respect. No action needed, just noting.

**19. Module docstring references "Sprint 34 contract" and "Sprint 35" — could be confusing**
The module docstring says "Sprint 35 Open Bandit OPE stack" but repeatedly references "Sprint 34 contract." Consider a brief note clarifying the relationship (Sprint 34 specifies the contract, Sprint 35 implements it).

**20. `GateReport.all_passed()` could use a more compact form**
```python
def all_passed(self) -> bool:
    return all(
        getattr(self, f).passed
        for f in ("null_control", "ess", "zero_support", "propensity_sanity", "dr_cross_check")
    )
```
This is marginally more DRY but also less explicit. The current explicit form is fine — it's just 5 fields and the explicit listing catches accidental omissions at review time.

---

### Summary

| Category | Count |
|----------|-------|
| Potential bugs | 3 |
| Numerical stability | 2 |
| Type safety / API | 3 |
| Design | 3 |
| Test gaps | 4 |
| Performance | 2 |
| Minor / style | 3 |

The implementation is solid overall. The statistical formulas for SNIPW, DM, and DR are correct. The gate logic faithfully implements the contract. The most important items to address are:

1. **Item 1** (null control gate semantics) — verify this matches the contract intent
2. **Item 7** (use `Literal` for `estimator` parameter)
3. **Items 13-14** (test coverage gaps for heterogeneous propensities and varying seeds)
4. **Item 9** (test-only parameter in production API)

---

datablogin:
# Claude Code Review

## Review Feedback

Let me proceed with the review based on my thorough reading of the code. I have enough context from the source files.

---

## PR Review: Sprint 35.B — Open Bandit SNIPW/DM/DR Stack and Section 7 Gates

### Overall Assessment

This is a well-structured, well-documented module. The statistical implementations are correct, the test coverage is thorough, and the code follows the project's conventions. Below are issues I found, organized by severity.

---

### Bugs / Correctness Issues

**1. `null_control_gate` evaluates SNIPW against *permuted* data but compares to mean of *permuted* rewards — the 5% band semantics may be subtly wrong**

`open_bandit.py:660-693` — The gate computes `mu_null = permuted["reward"].mean()` and then evaluates each strategy's SNIPW on the permuted data. The threshold is `band_multiplier * mu_null`. Under a uniform eval policy with uniform logging, SNIPW on permuted data collapses to the permuted reward mean, so the gate trivially passes. This is correct behavior for the null control — you're checking that no strategy can extract signal from noise.

However, the comparison `v <= threshold` where `threshold = band_multiplier * mu_null` uses the *raw reward mean* as the baseline, not the SNIPW-of-uniform as the baseline. For non-uniform eval policies, the SNIPW under permuted rewards can legitimately differ from `mu_null` due to position-reward correlation surviving the stratified permutation. The 5% band may be too tight for peaked policies on position-correlated data. This is more of a design consideration than a bug, but worth noting for when real OBD data lands.

**2. `null_control_gate` when `mu_null` is zero and `strategy_value_overrides` injects a zero value**

`open_bandit.py:691-694` — When `mu_null == 0.0`, `threshold = 1.05 * 0.0 = 0.0`. The check `all(v <= 0.0 for v in ...)` means any positive SNIPW from floating-point noise will fail. This is the intended "conservative" behavior per the comment, but any non-zero-reward permuted data would produce a nonzero `mu_null`, so in practice this only fires on all-zero reward slices. Acceptable but fragile.

**3. `n_positions` inference from `position.max() + 1` is fragile**

`open_bandit.py:668-669` — The code computes `n_positions = int(position.max() + 1)`, which assumes 0-indexed contiguous positions. The comment explicitly calls this out and defers to Issue A's smoke test. Good that it's documented, but this could silently produce wrong propensity clip floors if real OBD data has 1-indexed positions. Consider adding a validation check (e.g., assert `position.min() == 0`) rather than just a comment.

---

### Statistical / Numerical Issues

**4. ESS computation uses `1e-300` floor instead of checking for zero sum-of-squares**

`open_bandit.py:465` — `max(float((weights**2).sum()), 1e-300)` is a reasonable numerical guard, but `1e-300` is oddly chosen. If `weights.sum() > 0` (the outer condition), then `(weights**2).sum()` is also positive, so this guard never fires. The `1e-300` is dead code. Either remove the `max()` or document why it's needed.

**5. SNIPW `denom == 0.0` check uses exact float comparison**

`open_bandit.py:334` — `if denom == 0.0:` is fine here because `denom` is a sum of non-negative values (policy mass / clipped pscore), so zero only occurs when *all* weights are exactly zero. This is mathematically correct. No issue.

**6. DR estimator is not self-normalized**

`open_bandit.py:357-403` — The in-module DR uses the unnormalized form (Dudík et al. 2011), while SNIPW is self-normalized. This is fine for correctness — DR's bias reduction comes from the reward model, not from self-normalization — but the cross-check gate (7e) compares the two values with a 25% relative tolerance. Under high-variance weight scenarios, the unnormalized DR could diverge from SNIPW simply due to the normalization difference, not due to model misspecification. This could cause false positives in the cross-check gate. Worth a docstring note.

---

### Code Quality

**7. `evaluate_open_bandit_policy` returns `dict[str, Any]` — should be a typed dataclass or TypedDict**

`open_bandit.py:409-514` — The module already defines clean dataclasses for gate results, but the main evaluation function returns an untyped `dict[str, Any]`. This forces callers to use string-keyed access and defeats mypy. A `PolicyEvaluationResult` dataclass or TypedDict would be more consistent with the rest of the module and the project convention of using typed structures.

**8. `strategy_value_overrides` parameter is a test hook in production API**

`open_bandit.py:633` — The `null_control_gate` function has a `strategy_value_overrides` parameter that exists solely for testing. This leaks test infrastructure into the public API. A cleaner approach would be to test the gate's pass/fail logic by constructing bandit feedback with known rewards that deterministically produce above-band SNIPW values, or by making the SNIPW computation injectable. That said, the parameter is clearly documented as test-only and does no harm beyond API surface area.

**9. `__init__.py` `__all__` is getting large (50+ entries)**

`__init__.py:147-239` — The `__all__` list now has ~70 entries. It's maintained in roughly alphabetical order but the open_bandit additions are interleaved, which is correct. This isn't a problem now, but the benchmarks package may benefit from sub-namespace grouping in the future.

**10. `PropensitySchema` is a type alias, not a class — `test_propensity_schema_literal_type` tests nothing meaningful**

`test_open_bandit_ope.py:741-743` — `assert PropensitySchema is not None` only checks that the import succeeded. Since `Literal["conditional", "joint"]` is erased at runtime to just `typing.Literal`, this test provides no value beyond import smoke testing. The adjacent `test_schema_constants_stable` already covers the relevant invariants.

---

### Test Coverage Gaps

**11. No test for `compute_dr` with reward_hat that meaningfully biases the DM term**

The DR tests verify exact-residual and hand-computed cases, but don't test the scenario where `reward_hat` is systematically wrong (e.g., overestimates one action). This is the scenario where DR's doubly-robust property should shine — the IPW correction compensates for the biased DM term.

**12. No test for `evaluate_open_bandit_policy` with `estimator="dr"` in the integration tests**

`test_open_bandit_gates.py:61-74` — The integration test computes `compute_dr` directly but never calls `evaluate_open_bandit_policy(..., estimator="dr")`. This path is only tested in the unit tests' input validation section.

**13. No test for the `weight_cv` diagnostic**

`weight_cv` is computed in `evaluate_open_bandit_policy` but never verified for correctness — only checked for presence in the diagnostic dict via `test_diagnostic_keys_present`. A simple test with hand-computed weights would catch regressions in the CV formula.

**14. `null_control_gate` with empty `strategy_policies` is untested**

Calling `null_control_gate` with `strategy_policies={}` would return `passed=True` (vacuously true from `all()` on empty iterator). This may or may not be the intended behavior — it should be tested or guarded with a validation check.

---

### Minor / Style

**15. Redundant `float()` wrapping**

`open_bandit.py:505-513` — Several values are wrapped with `float()` despite already being floats (e.g., `float(policy_value)` when `compute_snipw` already returns `float`). This is harmless but adds noise.

**16. `peaked_policy` normalization is unnecessary given the construction**

`open_bandit.py:263-268` — The policy is constructed such that `peak_mass + (n_actions - 1) * remainder = 1.0` by definition. The normalization step at line 267-268 adds overhead (a full-array division) for a guarantee that already holds analytically. The comment says "numerical guard" but with IEEE 754 doubles this construction doesn't accumulate enough error to matter for n_actions up to millions.

**17. `generate_synthetic_bandit_feedback` docstring says "sampled uniformly over `n_actions * n_positions` logger actions" but the code samples `action` and `position` independently**

`open_bandit.py:144-145` vs `191-192` — The docstring implies joint uniform sampling over action-position pairs, but the code samples `action ~ Uniform(0, n_actions)` and `position ~ Uniform(0, n_positions)` independently. The result is the same distribution (uniform over the product space), but the wording "over `n_actions * n_positions` logger actions" is slightly misleading since actions and positions are independent random variables here.

---

### Summary

The implementation is solid. The estimators are mathematically correct, the gate logic follows the contract spec, and test coverage is strong. The main actionable items are:

1. **Consider a TypedDict or dataclass for `evaluate_open_bandit_policy`'s return value** — highest-impact improvement for type safety
2. **Add a position-index validation** in `null_control_gate` (assert `position.min() == 0`) rather than relying on a comment
3. **Add `weight_cv` correctness test** — currently untested
4. **Document the normalized-vs-unnormalized asymmetry** between SNIPW and DR in the cross-check gate docstring
5. **Consider validating non-empty `strategy_policies`** in `null_control_gate`

---

datablogin:
# Claude Code Review

## Review Feedback

Let me proceed with the review based on my thorough reading of the code.

---

## PR Review: Sprint 35.B — Open Bandit SNIPW/DM/DR Stack and Section 7 Gates

Overall this is a well-structured, cleanly written module. The statistical implementations are correct, the test coverage is thorough, and the code follows project conventions. Below are findings organized by severity.

---

### Potential Bugs / Correctness Issues

**1. Null control gate: `mu_null` is computed from permuted rewards, but the threshold check is asymmetric**
`open_bandit.py:698-699` — The gate checks `v <= threshold` where `threshold = band_multiplier * mu_null`. When rewards are non-negative (Bernoulli), this works. But if rewards could be negative, a strategy with a *very negative* value would pass trivially even if it diverges wildly from `mu_null`. The contract says "5% relative band", but the implementation is a one-sided upper bound, not `abs(v - mu_null) <= band * mu_null`. If the contract intends one-sided, this is fine — but worth documenting the design choice explicitly.

**2. `null_control_gate` checks 0-indexing only in the gate, not in `permute_rewards_stratified`**
`open_bandit.py:669` — The 0-indexing validation happens inside `null_control_gate` but `permute_rewards_stratified` is also a public function that could be called independently. A caller using `permute_rewards_stratified` directly with 1-indexed positions would silently succeed but produce semantically incorrect strata counts downstream. Consider moving the validation into `permute_rewards_stratified` or at least documenting this precondition.

**3. `n_positions` inferred from `position.max() + 1` assumes contiguous positions**
`open_bandit.py:674` — If positions are `{0, 2}` (0-indexed but with a gap), `n_positions` would be 3, not 2. The 0-indexing check catches the non-zero-min case, but doesn't catch gaps. This could silently mis-scale the propensity clip floor. The docstring says "0-indexed contiguous integers" but there's no validation that positions are actually contiguous.

**4. ESS computed twice in the SNIPW path**
`open_bandit.py:459-467` then `open_bandit.py:486-488` — When `estimator="snipw"`, `evaluate_open_bandit_policy` computes clipped weights and all diagnostics, then calls `compute_snipw` which recomputes clipped weights and importance weights from scratch. This is redundant work, not a bug, but on large datasets it doubles the cost of the primary path.

---

### Statistical / Numerical Considerations

**5. SNIPW denom check uses exact `== 0.0` float comparison**
`open_bandit.py:334` — `if denom == 0.0` is safe here because weights are non-negative (ratio of non-negative policy mass over positive clipped pscore), so `sum == 0` only when all weights are literally zero. But a comment explaining why exact float comparison is safe would aid maintainability. The existing comment explains the *behavior* but not the *safety* of the float comparison.

**6. `weight_cv` uses population std (`ddof=0`) rather than sample std**
`open_bandit.py:469` — With `ddof=0`, the CV is slightly biased downward for small samples. This is fine for a diagnostic, but deviates from the convention that `ddof=1` is the default in most statistical software. Worth noting whether the contract specifies population or sample std. If users will compare this CV against thresholds from literature, `ddof=1` is more standard.

**7. DR formula doesn't self-normalize the IPW correction**
`open_bandit.py:395-403` — The DR implementation uses vanilla (non-self-normalized) IPW for the correction term. This is mathematically correct per Dudík et al. 2011, but SNIPW uses self-normalization. When comparing DR and SNIPW in the 7e cross-check gate, the difference between self-normalized and non-normalized formulations could itself cause divergence, which would be a false positive from the gate. The docstring mentions this but the gate tolerance (25%) should be validated empirically to account for this structural difference.

---

### Design / API

**8. `evaluate_open_bandit_policy` returns `dict[str, Any]` instead of a typed dataclass**
The gate results are properly typed dataclasses, but the primary evaluation function returns an untyped dict. This is consistent with the contract's Section 4d spec but creates a type-safety gap — callers access string keys without compile-time checking. A `PolicyEvaluationResult` dataclass would be more consistent with the gate result pattern and would catch key typos at type-check time.

**9. `strategy_value_overrides` is a test-only parameter on a production function**
`open_bandit.py:633` — The `null_control_gate` function accepts `strategy_value_overrides` solely for test determinism. This is a testing backdoor in the production API. Consider either: (a) removing it and having tests inject determinism via the permutation seed + synthetic data design, or (b) making it a private `_strategy_value_overrides` parameter to signal it's internal.

**10. `GateReport.all_passed()` could use a loop over fields for maintainability**
`open_bandit.py:613-620` — If a sixth gate is added, `all_passed()` must be manually updated. Using `dataclasses.fields()` to iterate over all fields that have a `passed` attribute would be more robust:
```python
def all_passed(self) -> bool:
    return all(
        getattr(self, f.name).passed
        for f in fields(self)
        if hasattr(getattr(self, f.name), 'passed')
    )
```

---

### Test Coverage

**11. No test for `n_actions=1` edge case**
With a single action, the peaked policy lower bound (`1/n_actions = 1.0`) means only `peak_mass=1.0` is valid, which is identical to `degenerate_policy`. The estimators should still work but this edge case (common in A/B testing) is untested.

**12. No test for DR with clipped propensities**
The DR tests all use propensities well above the clip floor. A test where `pscore` values are below `min_propensity_clip` would verify that DR's correction term handles clipped weights correctly (the correction should be dampened by clipping).

**13. No negative test for `run_section_7_gates` with missing `position` key**
The bundle function delegates to `null_control_gate` which delegates to `permute_rewards_stratified`, which raises `KeyError` on missing position. But the error path through `run_section_7_gates` isn't directly tested.

**14. Integration test ESS values are implicitly coupled to synthetic data**
`test_open_bandit_gates.py:70-71` — The ESS values are computed from `evaluate_open_bandit_policy` on uniform-logger/uniform-policy synthetic data where all weights are exactly 1.0. This means `ess == n_rounds == 5000`, which easily clears the floor of `max(1000, 5000/100) = 1000`. The test works but doesn't exercise the ESS gate near its boundary. A test with `n_rounds` just above and just below the floor crossing point would be more robust.

---

### Code Quality

**15. `__init__.py` `__all__` ordering is inconsistent**
The `__all__` list mixes uppercase constants, classes, and functions but doesn't follow a consistent sort within each category. Some entries appear to be alphabetically sorted globally, but `HILLSTROM_VALID_STRATEGIES` appears after `HILLSTROM_PRIMARY_PROPENSITY`. This is a pre-existing issue amplified by the new additions.

**16. Imports in `__init__.py` are eagerly loaded**
All 30+ symbols from `open_bandit` are imported at the top level of `benchmarks/__init__.py`. This means `import causal_optimizer.benchmarks` now triggers numpy import even for callers who only need the SCM benchmarks. Consider lazy imports or splitting into sub-packages if startup time becomes an issue (though numpy is likely already imported by other paths).

**17. Module docstring is thorough but could separate the public API listing**
`open_bandit.py:1-62` — The 62-line module docstring is comprehensive but the inline public API list duplicates what `__all__` would provide. The narrative sections (Sections 4d, 5, 6, 7 explanations) are valuable; the API listing is noise.

---

### Minor Nits

- `open_bandit.py:88` — `PropensitySchema = Literal["conditional", "joint"]` is a type alias, not a runtime constant. Consider adding `from typing import TypeAlias` and annotating as `PropensitySchema: TypeAlias = Literal[...]` for clarity (Python 3.10+).
- `open_bandit.py:90` — `OBP_VERSION_PLACEHOLDER` is 85 characters of prose as a default value. The constant itself is fine but the em-dash character (`—`) in the string may cause encoding issues in some CI environments.
- `test_open_bandit_ope.py:76` — Line exceeds 100 chars (the ruff limit). Check with `ruff check`.

---

### Summary

| Category | Finding Count |
|----------|:---:|
| Potential bugs | 4 |
| Statistical/numerical | 3 |
| Design/API | 3 |
| Test coverage gaps | 4 |
| Code quality | 3 |
| Minor nits | 3 |

**Verdict**: This is solid work. The estimator implementations are mathematically correct, the gate structure is well-designed with good separation of concerns, and the test suite is comprehensive. The main items I'd want addressed before merge are #2 (0-indexing validation scope), #3 (contiguous position validation), and #9 (test-only parameter in production API). The rest are improvements that could land in follow-up PRs.

---

datablogin:
@greptileai

PR #189, opened 2026-04-20
Sprint 35.A: Open Bandit DomainAdapter and Men/Random smoke test

### LINKED ISSUE #185: Sprint 35.A: implement the Open Bandit DomainAdapter and smoke test

datablogin:
## Summary

Implements the Sprint 34 Open Bandit contract Sections 4 and 8 for Track A of issue #185. Ships a new `BanditLogAdapter` that does **not** subclass `MarketingLogAdapter` (contract Section 4a), parameterized as a six-variable contextual item-scoring policy (`tau`, `eps`, three context-feature weights, `position_handling_flag`) and evaluated with a minimal in-house SNIPW-style estimator over a logged multi-action bandit-feedback dict.

Also adds an optional `bandit` extra pinning `obp==0.4.1` (Apache 2.0) so the Open Bandit Pipeline data loader is available via `BanditLogAdapter.from_obp` while the core install stays OBP-free. The adapter fails fast with an actionable error when OBP is missing (contract Section 8c).

Track A scope boundary: the adapter ships the Section 4 interface surface and a narrow first-pass SNIPW-style `policy_value` calculation so the adapter is reviewable end-to-end without blocking on the OPE stack. Track B (issue #186) owns the production OPE stack (OBP-backed SNIPW / DM / DR + Section 7 support gates + DR/SNIPW cross-check).

Closes #185.

## Section 10.A smoke-test findings (pinned in the adapter docstring)

1. **Row count.** The `obp==0.4.1` wheel bundles a 10,000-row sample of the Men/Random slice. The full ~452,949-row slice (Saito et al. 2021 Table 1) requires passing `data_path=` to `BanditLogAdapter.from_obp`; Issue C will exercise that path.
2. **`action_prob` / `pscore` schema.** Confirmed as **conditional `P(item | position)`**. Empirical mean `0.029412` matches `1/n_items = 1/34` to floating-point precision; not joint `1/(34*3) ≈ 0.0098`. Section 7d's propensity-mean sanity gate should therefore use the `1/n_items` target on Men/Random.
3. **Chosen three context features** (Section 4c):
   - `w_item_feature_0` — per-item continuous feature from `item_context.csv` (range ≈ `[-0.73, 0.75]`).
   - `w_user_item_affinity` — per-row per-candidate affinity lookup (34 columns in the raw log).
   - `w_item_popularity` — per-item log-normalized in-log appearance count, pre-computed at construction.

## Files touched

- `pyproject.toml` — new `bandit` optional extra, new `obp` pytest marker.
- `causal_optimizer/domain_adapters/bandit_log.py` (new) — `BanditLogAdapter` implementation.
- `causal_optimizer/domain_adapters/__init__.py` — export `BanditLogAdapter`.
- `tests/unit/test_bandit_log_adapter.py` (new) — 32 fast tests against synthetic bandit-feedback dicts.
- `tests/integration/test_bandit_log_adapter_smoke.py` (new) — 5 slow tests against the bundled OBP Men/Random sample.

## Test plan

- [x] `uv run pytest tests/unit/test_bandit_log_adapter.py` — 32 passed
- [x] `uv run pytest tests/integration/test_bandit_log_adapter_smoke.py -m slow` — 5 passed
- [x] `uv run pytest -m "not slow"` — 1201 passed, 24 skipped (full fast suite green)
- [x] `uv run ruff check .` — all checks passed
- [x] `uv run ruff format --check .` — no diff
- [x] `uv run mypy causal_optimizer/` — 63 source files clean (strict mode)

## Known caveats

- OBP 0.4.1's `pre_process()` calls `DataFrame.drop(..., 1)` with a positional `axis` argument that modern pandas rejects. The smoke test reads the bundled CSVs directly to bypass that pre-processor. Issue B / Issue C can switch to a patched loader later; the adapter boundary does not depend on `pre_process()`.
- Coverage reporting (`--cov`) fails in the local dev environment under `numpy==2.4.2` with "cannot load module more than once per process" — pre-existing on `main` for every numpy-touching test file, not a regression. All meaningful branches in `bandit_log.py` are exercised by explicit tests (validation error paths, both position flags, zero-active-rows fallback, invalid flag, narrow-context fallback, `from_obp` missing + success-delegation stub).

<h3>Greptile Summary</h3>

This PR ships `BanditLogAdapter`, a new `DomainAdapter` for logged multi-action bandit-feedback data implementing the Sprint 34 Open Bandit contract Sections 4 and 8. It adds a six-variable contextual item-scoring policy with a self-normalized IPW-style estimator, optional OBP data-loading via `from_obp`, and 37 tests (32 unit + 5 integration smoke tests against the bundled Men/Random slice).

Both concerns flagged in the previous review are resolved — the `pytestmark` at line 27 now correctly applies both `slow` and `obp` marks, and the `isinstance` guard in `from_obp` uses a real `TypeError` rather than a strippable `assert`.

Safe to merge — no P0/P1 issues found; both prior review concerns are resolved.

Both previous findings (missing obp marker and assert-based runtime guard) are correctly addressed. The SNIPW estimator, softmax, validation logic, and all edge-case paths are sound. The 37-test suite is comprehensive and the integration smoke test now applies the correct dual slow+obp pytestmark. No new blocking issues identified.

| Filename | Overview |
|----------|----------|
| causal_optimizer/domain_adapters/bandit_log.py | New BanditLogAdapter: thorough validation, numerically stable softmax, correct SNIPW estimator, and well-bounded diagnostic dict. No logic errors found. |
| tests/unit/test_bandit_log_adapter.py | 32 fast unit tests covering all validation paths, semantic invariants, position flag behavior, and the OBP lazy-import contract. Well-structured. |
| tests/integration/test_bandit_log_adapter_smoke.py | 5 smoke tests against bundled Men/Random slice; pytestmark now correctly applies both slow and obp markers (addresses previous review thread). |
| pyproject.toml | Adds bandit optional extra pinning obp>=0.4.1,<0.5 and registers the obp pytest marker. Upper-bound on obp protects against API churn. |
| causal_optimizer/domain_adapters/__init__.py | Exports BanditLogAdapter; alphabetically ordered and __all__ updated correctly. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant BanditLogAdapter
    participant OBP as OpenBanditDataset (optional)
    participant SNIPW as SNIPW Estimator

    alt Direct construction
        Caller->>BanditLogAdapter: __init__(bandit_feedback, seed)
        BanditLogAdapter->>BanditLogAdapter: _validate_feedback()
        BanditLogAdapter->>BanditLogAdapter: pre-compute item_feature_0, affinity, popularity
    else Via OBP loader
        Caller->>BanditLogAdapter: from_obp(campaign, behavior_policy, data_path)
        BanditLogAdapter->>OBP: lazy import obp.dataset (fails fast if missing)
        OBP-->>BanditLogAdapter: OpenBanditDataset
        BanditLogAdapter->>OBP: obtain_batch_bandit_feedback()
        OBP-->>BanditLogAdapter: bandit_feedback dict
        BanditLogAdapter->>BanditLogAdapter: __init__(bandit_feedback, seed)
    end

    Caller->>BanditLogAdapter: run_experiment(parameters)
    BanditLogAdapter->>BanditLogAdapter: apply position mask
    BanditLogAdapter->>BanditLogAdapter: score = item + affinity + popularity terms
    BanditLogAdapter->>BanditLogAdapter: policy = softmax(score/tau) mixed with eps*uniform
    BanditLogAdapter->>SNIPW: weights = pi_e(a|x) / pi_b(a|x)
    SNIPW-->>BanditLogAdapter: policy_value = sum(w*r) / sum(w)
    BanditLogAdapter-->>Caller: diagnostic dict (6 float keys)
```

<sub>Reviews (3): Last reviewed commit: ["address claude review feedback (claudelo..."](https://github.com/datablogin/causal-optimizer/commit/64d89ef834df4b6398814f5a53f9379e131e20d0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28929664)</sub>

### DIFF
--- a/causal_optimizer/domain_adapters/__init__.py
+++ b/causal_optimizer/domain_adapters/__init__.py
@@ -1,5 +1,6 @@
 """Domain adapters — plug in different experiment domains."""
 
+from causal_optimizer.domain_adapters.bandit_log import BanditLogAdapter
 from causal_optimizer.domain_adapters.base import DomainAdapter
 from causal_optimizer.domain_adapters.energy_load import EnergyLoadAdapter
 from causal_optimizer.domain_adapters.marketing import MarketingAdapter
@@ -7,6 +8,7 @@ from causal_optimizer.domain_adapters.marketing_logs import MarketingLogAdapter
 from causal_optimizer.domain_adapters.ml_training import MLTrainingAdapter
 
 __all__ = [
+    "BanditLogAdapter",
     "DomainAdapter",
     "EnergyLoadAdapter",
     "MarketingAdapter",
--- a/causal_optimizer/domain_adapters/bandit_log.py
+++ b/causal_optimizer/domain_adapters/bandit_log.py
@@ -563,3 +563,15 @@ class BanditLogAdapter(DomainAdapter):
                 f"bandit_feedback['action'] values must be in [0, {n_actions}); "
                 f"found range [{int(action.min())}, {int(action.max())}]"
             )
+
+        # The adapter only interprets ``position == 0`` as "position 1"
+        # (mirroring OBP's ``rankdata`` convention). Negative position
+        # values would silently distort the ``position_1_only`` subset
+        # without raising, so reject them at construction.
+        position = np.asarray(bandit_feedback["position"])
+        if position.min() < 0:
+            raise ValueError(
+                "bandit_feedback['position'] values must be non-negative; "
+                f"found min={int(position.min())}. OBP stores positions "
+                "as 0-indexed integers after ``rankdata(..., 'dense')``."
+            )
--- a/causal_optimizer/domain_adapters/bandit_log.py
+++ b/causal_optimizer/domain_adapters/bandit_log.py
@@ -1,0 +1,523 @@
+"""Open Bandit ``DomainAdapter`` for logged multi-action policy data.
+
+Implements the Sprint 34 Open Bandit contract (Sections 4 and 8) for the
+ZOZOTOWN Men / uniform-random slice. Parameterizes a contextual
+item-scoring policy in a six-variable search space (``tau`` softmax
+temperature, ``eps`` exploration epsilon, three context-feature weights,
+a ``position_handling_flag``) and evaluates it against a logged
+bandit-feedback dict with a minimal in-house SNIPW-style estimator.
+
+**Track A scope.** This adapter ships the Section 4 interface surface
+and a narrow first-pass SNIPW-style ``policy_value`` calculation so the
+adapter is reviewable end-to-end without blocking on the OPE stack.
+Track B (Sprint 35.B, issue #186) owns the production OPE stack:
+the OBP-backed SNIPW / DM / DR estimators, the Section 7 support gates,
+and the cross-estimator cross-check.
+
+Sprint 35.A smoke-test findings (Sprint 34 contract Section 10.A)
+-----------------------------------------------------------------
+
+1. **Row count.** The ``obp==0.4.1`` wheel bundles a 10,000-row sample
+   of the Men/Random slice at
+   ``site-packages/obp/dataset/obd/random/men/men.csv``. The full
+   ~452,949-row slice from Saito et al. 2021 Table 1 must be loaded
+   separately by passing ``data_path=`` to :meth:`from_obp`; Issue C
+   will exercise that path, Issue A's smoke test pins only the bundled
+   sample.
+
+2. **``action_prob`` / ``pscore`` schema.** Confirmed as **conditional
+   ``P(item | position)``**. The empirical mean of
+   ``propensity_score`` on the bundled slice is ``0.0294117...``, which
+   matches ``1/n_items = 1/34`` to floating-point precision. It does
+   not match joint ``P(item, position) = 1/(n_items * n_positions) =
+   1/102 ≈ 0.0098``. All three positions report the same constant
+   propensity, consistent with the conditional interpretation under
+   uniform-random logging. Section 7d's propensity-mean sanity gate
+   should therefore use the ``1/n_items`` target on Men/Random.
+
+3. **Chosen context features (three).** The adapter's search space
+   pins three context-feature weights:
+
+   - ``w_item_feature_0`` — multiplies the per-item continuous
+     ``item_feature_0`` column from ``item_context.csv``
+     (range ≈ ``[-0.73, 0.75]`` on Men/Random).
+   - ``w_user_item_affinity`` — multiplies the per-row, per-candidate
+     ``user-item_affinity_<k>`` lookup (34 columns, one per item id).
+   - ``w_item_popularity`` — multiplies the per-item log-normalized
+     in-log appearance count, pre-computed at adapter construction
+     from ``bandit_feedback["action"]`` so the score is deterministic
+     across policy evaluations.
+
+   Rationale: these three features are the minimum honest subset that
+   (i) exposes per-item continuous signal, (ii) exposes per-(row, item)
+   heterogeneity, and (iii) anchors to an empirical popularity prior
+   without blowing up the search dimensionality. Each weight has a
+   continuous bound of ``[-3.0, 3.0]``.
+
+Public interface
+----------------
+
+:class:`BanditLogAdapter` takes a pre-materialized OBP-style
+bandit-feedback dict directly (the common case for unit tests and
+downstream OPE harnesses). Use :meth:`BanditLogAdapter.from_obp` to
+load the Men/Random slice through OBP; this path requires the optional
+``bandit`` extra (``uv sync --extra bandit``) and fails fast with an
+actionable error if ``obp`` is missing.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+from causal_optimizer.domain_adapters.base import DomainAdapter
+from causal_optimizer.types import SearchSpace, Variable, VariableType
+
+if TYPE_CHECKING:
+    from causal_optimizer.types import CausalGraph
+
+__all__ = ["BanditLogAdapter"]
+
+
+# ── Constants ────────────────────────────────────────────────────────
+
+_REQUIRED_FEEDBACK_KEYS: frozenset[str] = frozenset(
+    {
+        "n_rounds",
+        "n_actions",
+        "action",
+        "position",
+        "reward",
+        "pscore",
+        "context",
+        "action_context",
+    }
+)
+
+_CONTEXT_WEIGHT_NAMES: tuple[str, ...] = (
+    "w_item_feature_0",
+    "w_user_item_affinity",
+    "w_item_popularity",
+)
+
+_POSITION_HANDLING_CHOICES: tuple[str, ...] = ("marginalize", "position_1_only")
+
+# Search-space bounds — see Sprint 34 contract Section 4c.
+_TAU_LOWER: float = 0.1
+_TAU_UPPER: float = 10.0
+_EPS_LOWER: float = 0.0
+_EPS_UPPER: float = 0.5
+_WEIGHT_LOWER: float = -3.0
+_WEIGHT_UPPER: float = 3.0
+
+# Effective-action mass threshold: n_effective_actions is the smallest
+# k such that the top-k actions' combined policy mass exceeds this
+# threshold. Pinned at 0.95 per Sprint 34 contract Section 4d.
+_EFFECTIVE_ACTIONS_MASS_THRESHOLD: float = 0.95
+
+# Minimum tau used inside the softmax. If a caller passes ``tau=0`` (or
+# numerically close) the adapter floors the temperature to keep the
+# softmax numerically stable; the search space bounds already forbid
+# zero, this is a defence against numerical edge cases inside the
+# optimizer.
+_TAU_FLOOR: float = 1e-6
+
+
+# ── Adapter ──────────────────────────────────────────────────────────
+
+
+class BanditLogAdapter(DomainAdapter):
+    """``DomainAdapter`` for logged multi-action bandit-feedback data.
+
+    Parameters
+    ----------
+    bandit_feedback:
+        OBP-shaped ``dict`` with the keys ``n_rounds``, ``n_actions``,
+        ``action``, ``position``, ``reward``, ``pscore``, ``context``,
+        and ``action_context``. Use :meth:`from_obp` to build this
+        dict from OBP's ``OpenBanditDataset``; pass one directly in
+        unit tests and when consuming a pre-cached slice.
+    seed:
+        Reproducibility seed. Accepted for API consistency with
+        other adapters — evaluation is fully deterministic given
+        the bandit-feedback dict and parameters, so this value is
+        forwarded to :class:`numpy.random.Generator` only if future
+        stochastic diagnostics are added.
+
+    Notes
+    -----
+    Sprint 34 contract Section 4a forbids subclassing
+    ``MarketingLogAdapter``. This class inherits directly from
+    :class:`DomainAdapter` and exposes the Section 4d diagnostic dict
+    (``policy_value``, ``ess``, ``weight_cv``, ``max_weight``,
+    ``zero_support_fraction``, ``n_effective_actions``).
+    """
+
+    def __init__(
+        self,
+        *,
+        bandit_feedback: dict[str, Any],
+        seed: int | None = None,
+    ) -> None:
+        self._validate_feedback(bandit_feedback)
+        self._seed = seed
+
+        self._n_rounds: int = int(bandit_feedback["n_rounds"])
+        self._n_actions: int = int(bandit_feedback["n_actions"])
+        self._action: np.ndarray = np.asarray(bandit_feedback["action"], dtype=np.int64)
+        self._position: np.ndarray = np.asarray(bandit_feedback["position"], dtype=np.int64)
+        self._reward: np.ndarray = np.asarray(bandit_feedback["reward"], dtype=np.int64)
+        self._pscore: np.ndarray = np.asarray(bandit_feedback["pscore"], dtype=float)
+        self._context: np.ndarray = np.asarray(bandit_feedback["context"], dtype=float)
+        self._action_context: np.ndarray = np.asarray(
+            bandit_feedback["action_context"], dtype=float
+        )
+
+        # Pre-compute per-item score components once. These are
+        # independent of the policy parameters, so hoisting them out of
+        # run_experiment keeps the optimizer loop cheap.
+        self._item_feature_0: np.ndarray = self._action_context[:, 0].astype(float)
+
+        # Per-row, per-candidate affinity matrix. The raw log's
+        # ``user-item_affinity_<k>`` columns are per-row lookups for
+        # candidate item k; ``context`` is therefore expected to have
+        # one column per action. When ``context`` has fewer columns the
+        # adapter falls back to a zero-affinity surface (still valid,
+        # the search space's ``w_user_item_affinity`` weight then has
+        # no effect — useful for small synthetic fixtures).
+        if self._context.shape[1] >= self._n_actions:
+            self._affinity: np.ndarray = self._context[:, : self._n_actions]
+        else:
+            self._affinity = np.zeros((self._n_rounds, self._n_actions), dtype=float)
+
+        # Per-item popularity prior (log-normalized appearance count).
+        # Computed once at construction from ``action`` so the prior is
+        # deterministic across policy evaluations.
+        counts = np.bincount(self._action, minlength=self._n_actions).astype(float)
+        raw_max = float(counts.max())
+        max_count = raw_max if raw_max > 0.0 else 1.0
+        self._item_popularity: np.ndarray = np.log1p(counts) / np.log1p(max_count)
+
+    # ── Constructor from OBP ──────────────────────────────────────────
+
+    @classmethod
+    def from_obp(
+        cls,
+        *,
+        campaign: str = "men",
+        behavior_policy: str = "random",
+        data_path: Any | None = None,
+        seed: int | None = None,
+    ) -> BanditLogAdapter:
+        """Build an adapter from an OBP ``OpenBanditDataset``.
+
+        Loads the requested campaign / logger slice via OBP, extracts
+        the bandit-feedback dict, and returns a ready-to-use adapter.
+        Requires the optional ``bandit`` extra (``uv sync --extra
+        bandit``); raises a clear :class:`ImportError` if the extra is
+        not installed.
+
+        Parameters
+        ----------
+        campaign:
+            ZOZOTOWN campaign name. Must be one of ``"all"``, ``"men"``,
+            or ``"women"``. Defaults to ``"men"`` per the Sprint 34
+            first-slice decision.
+        behavior_policy:
+            Logging policy. Must be ``"random"`` or ``"bts"``. Defaults
+            to ``"random"`` per the Sprint 34 first-slice decision.
+        data_path:
+            Optional ``pathlib.Path`` to the full released dataset.
+            When ``None`` the OBP bundled small-sized sample is used
+            (10,000 rows on Men/Random as of obp 0.4.1).
+        seed:
+            Reproducibility seed; forwarded to the adapter constructor.
+        """
+        try:
+            # Inside-function import is deliberate: the ``obp`` package
+            # is the optional ``bandit`` extra, and keeping this import
+            # lazy is what lets the core adapter module stay importable
+            # without the extra installed.
+            from obp.dataset import OpenBanditDataset
+        except ImportError as exc:  # pragma: no cover - exercised by monkeypatched test
+            msg = (
+                "BanditLogAdapter.from_obp requires the 'bandit' extra. "
+                "Install it with: uv sync --extra bandit  "
+                "(or: pip install 'causal-optimizer[bandit]')."
+            )
+            raise ImportError(msg) from exc
+
+        dataset = OpenBanditDataset(
+            behavior_policy=behavior_policy,
+            campaign=campaign,
+            data_path=data_path,
+        )
+        bandit_feedback = dataset.obtain_batch_bandit_feedback()
+        # OBP's obtain_batch_bandit_feedback returns a dict with exactly
+        # the keys we require; validate defensively so a future OBP
+        # upgrade does not silently break us.
+        assert isinstance(bandit_feedback, dict)
+        return cls(bandit_feedback=bandit_feedback, seed=seed)
+
+    # ── DomainAdapter interface ────────────────────────────────────────
+
+    def get_search_space(self) -> SearchSpace:
+        variables: list[Variable] = [
+            Variable(
+                name="tau",
+                variable_type=VariableType.CONTINUOUS,
+                lower=_TAU_LOWER,
+                upper=_TAU_UPPER,
+            ),
+            Variable(
+                name="eps",
+                variable_type=VariableType.CONTINUOUS,
+                lower=_EPS_LOWER,
+                upper=_EPS_UPPER,
+            ),
+        ]
+        for name in _CONTEXT_WEIGHT_NAMES:
+            variables.append(
+                Variable(
+                    name=name,
+                    variable_type=VariableType.CONTINUOUS,
+                    lower=_WEIGHT_LOWER,
+                    upper=_WEIGHT_UPPER,
+                )
+            )
+        variables.append(
+            Variable(
+                name="position_handling_flag",
+                variable_type=VariableType.CATEGORICAL,
+                choices=list(_POSITION_HANDLING_CHOICES),
+            )
+        )
+        return SearchSpace(variables=variables)
+
+    def run_experiment(self, parameters: dict[str, Any]) -> dict[str, float]:
+        tau = float(parameters.get("tau", 1.0))
+        eps = float(parameters.get("eps", 0.0))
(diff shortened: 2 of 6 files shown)


datablogin:
@greptileai

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the codebase patterns and the PR changes. Here's my review.

---

## PR Review: Sprint 35.A Open Bandit Adapter

Overall this is a well-structured PR. The adapter follows the existing `DomainAdapter` contract closely, the SNIPW estimator is statistically sound, and the test coverage is thorough. A few issues worth addressing:

### Bugs / Correctness Issues

**1. `reward` stored as `int64` but used as float — potential silent truncation**
`bandit_log.py:170`

```python
self._reward: np.ndarray = np.asarray(bandit_feedback["reward"], dtype=np.int64)
```

Then at line 361:
```python
active_reward = self._reward[mask].astype(float)
```

Storing as `int64` then converting to `float` on every `run_experiment` call is fine for binary {0,1} rewards, but it's an unnecessary conversion per call. More importantly, the validation at line 509-515 accepts any values that are *numerically* 0.0 or 1.0 (including `0.0` and `1.0` floats), but `np.asarray(..., dtype=np.int64)` will silently truncate float inputs like `0.999` to `0`. Consider storing as `float64` directly (consistent with `pscore`) and dropping the per-call `.astype(float)`.

**2. `__init__.py` eagerly imports `BanditLogAdapter` — breaks without numpy**
`domain_adapters/__init__.py:3`

The `BanditLogAdapter` module imports `numpy` at module level (line 72 of `bandit_log.py`), which is fine since numpy is a core dependency. However, the `from_obp` classmethod's lazy import of `obp` is correctly deferred. No actual bug here, but worth noting that the `__init__.py` eagerly loads the entire adapter class including all its constants. This is consistent with how the other adapters are handled, so this is fine.

### Statistical / Numerical Issues

**3. `weight_cv` uses `ddof=0` vs `ddof=1` in `MarketingLogAdapter`**
`bandit_log.py:387` vs `marketing_logs.py:347`

```python
# BanditLogAdapter (ddof=0)
weight_cv = float(positive.std(ddof=0) / positive.mean()) if positive.size > 1 else 0.0

# MarketingLogAdapter (ddof=1)
weight_cv = float(positive_weights.std(ddof=1) / positive_weights.mean())
```

The comment justifies `ddof=0` as "population std so the CV is well defined even at tiny n_active," which is reasonable. But this creates an inconsistency between the two adapters' diagnostic semantics. Users comparing `weight_cv` across adapters will get subtly different values at small sample sizes. Consider documenting which convention each adapter uses, or standardizing on one.

**4. Potential for SNIPW policy_value > 1.0 (or < 0.0)**
`bandit_log.py:374`

Self-normalized IPW does *not* guarantee the estimate stays in [0, 1] even for binary rewards. With skewed importance weights, SNIPW can produce estimates outside [0,1] (though it's rare with well-supported policies). The test at line 195 asserts `0.0 <= pv <= 1.0` — this could be a flaky test under adversarial parameter configurations. The `test_uniform_policy_has_policy_value_near_logged_ctr` test (with eps=0.5, zero weights) is safe, but sharp policies with eps=0 and extreme weights could theoretically violate the bound.

The `n_active == 0` early return correctly returns `policy_value=0.0`, so that edge case is fine. Consider whether a clamp or at minimum a comment explaining the theoretical possibility would be appropriate.

**5. `scaled` array is modified in-place**
`bandit_log.py:352`

```python
scaled = scores / safe_tau
scaled -= scaled.max(axis=1, keepdims=True)  # numerical stability
```

`scores / safe_tau` creates a new array, so the in-place subtraction is safe here. However, `scores` itself is computed from `item_term + pop_term + affinity_term` which *could* share memory with `self._affinity[mask]` in some numpy versions if the broadcasting happened to produce a view. In practice, the addition always creates a new array, so this is safe. Just flagging for awareness.

### Design / API Issues

**6. `n_effective_actions` returns 1.0 in the pessimistic early-return but the test asserts `>= 1.0`**
`bandit_log.py:339`

The pessimistic early return sets `n_effective_actions: 1.0`, which is semantically correct (a single action is the floor). But the `n_effective_actions` computation for the non-empty case (line 415) returns a float average. Good — this is consistent.

**7. `from_obp` doesn't validate `campaign` or `behavior_policy` values**
`bandit_log.py:204-269`

The docstring says `campaign` must be `"all"`, `"men"`, or `"women"` and `behavior_policy` must be `"random"` or `"bts"`, but the method doesn't validate these before passing them to `OpenBanditDataset`. OBP will raise its own error, but it would be more user-friendly to validate up front with a clear error message, consistent with the adapter's fail-fast philosophy. This is minor — OBP's error messages are adequate.

**8. `data_path` typed as `Any | None` rather than `Path | str | None`**
`bandit_log.py:210`

Using `Any` loses type safety. `pathlib.Path | str | None` would be more precise and still compatible with what `OpenBanditDataset` accepts.

### Performance

**9. `sorted_mass` / `cum_mass` allocation on every call**
`bandit_log.py:402-415`

The `n_effective_actions` computation creates three intermediate arrays (`sorted_mass`, `cum_mass`, `crosses`) of shape `(n_active, n_actions)` on every `run_experiment` call. For the 10K-row / 34-action dataset this is ~1MB per call — negligible. For the full 452K-row dataset this would be ~60MB per call, which is still fine but worth noting for Track B when the full slice lands.

**10. `item_term` and `pop_term` broadcast unnecessarily every call**
`bandit_log.py:344-345`

```python
item_term = w_item * self._item_feature_0[None, :]
pop_term = w_popularity * self._item_popularity[None, :]
```

These are `(1, n_actions)` arrays that get broadcast to `(n_active, n_actions)` during addition with `affinity_term`. The shapes are fine, but since `item_feature_0` and `item_popularity` don't change across calls, you could cache their `[None, :]` views in `__init__`. Minor optimization — not actionable for this PR.

### Test Coverage

**11. Missing test: `eps=0.0` with uniform scores**
When `eps=0.0` and all feature weights are zero, the softmax over identical scores produces a uniform policy. This is an important edge case that verifies `softmax(0/tau) = uniform` — it's implicitly tested via `test_uniform_policy_has_policy_value_near_logged_ctr` (which uses `eps=0.5`), but a dedicated test with `eps=0.0` would catch any softmax numerical issues with all-equal scores.

**12. No test for negative feature weights**
The search space allows `w_*` in `[-3, 3]`, but all tests use non-negative weights. A test with negative weights would exercise the "penalize popular items" or "penalize high-affinity items" path.

**13. Integration test `_load_men_random_raw` uses `os.path.dirname` instead of `Path.parent`**
`test_bandit_log_adapter_smoke.py:49`

```python
root = Path(os.path.dirname(obp.__file__)) / "dataset" / "obd" / "random" / "men"
```

Could be simplified to `Path(obp.__file__).parent / ...`. Minor style nit.

**14. Integration test doesn't verify `item_id` is 0-indexed**
`test_bandit_log_adapter_smoke.py:139`

```python
action = data["item_id"].to_numpy().astype(int)
```

If `item_id` in the raw CSV is 1-indexed (or has gaps), this would produce out-of-range action indices. The adapter's validation at line 526-531 would catch `action.max() >= n_actions`, but the smoke test doesn't explicitly verify the 0-indexing assumption. The validator *does* check this, so this is safe but could be more explicit.

### Code Quality

**15. Excellent docstring and module-level documentation.** The module docstring at lines 1-66 is exceptionally thorough — it documents the scope boundary (Track A vs Track B), the three smoke-test findings, and the public interface. This is a model for how domain adapters should be documented.

**16. Validation is thorough and actionable.** The `_validate_feedback` method provides clear, specific error messages for every failure mode. The separation between construction-time validation and runtime validation (in `run_experiment`) is well done.

**17. Clean separation of concerns.** Pre-computing `_item_feature_0`, `_affinity`, and `_item_popularity` in `__init__` keeps `run_experiment` focused on the policy evaluation logic.

### Summary

| Category | Finding | Severity |
|----------|---------|----------|
| Consistency | `weight_cv` ddof=0 vs ddof=1 across adapters | Medium |
| Type safety | `data_path: Any` should be `Path \| str \| None` | Low |
| Statistical | SNIPW not bounded to [0,1]; test could be flaky | Low |
| Robustness | `reward` stored as int64 then cast per call | Low |
| Test gap | No negative-weight or eps=0 uniform-score test | Low |
| Style | `os.path.dirname` vs `Path.parent` in smoke test | Nit |

The PR is well-scoped, well-tested, and follows existing adapter patterns closely. The SNIPW implementation is correct and numerically stable. The `ddof` inconsistency with `MarketingLogAdapter` is the most worth addressing before merge.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of both the new code and the existing patterns. Here's my review:

---

## PR Review: Sprint 35.A Open Bandit DomainAdapter

### Overall Assessment

This is a well-structured, thoughtful addition. The adapter is clean, well-documented, and the SNIPW estimator is implemented correctly. The test coverage is thorough with good semantic tests. A few issues worth addressing before merge:

---

### Bugs / Correctness Issues

**1. `run_experiment` mutates internal state via `scaled -= ...` (line 375)**

```python
scaled -= scaled.max(axis=1, keepdims=True)  # numerical stability
```

`scaled` is derived from `scores` which is a fresh array each call (from arithmetic ops), so this is safe — no mutation of `self._*` arrays. However, it's worth noting this is only safe because `scores = item_term + pop_term + affinity_term` creates a new array. If the scoring logic were ever refactored to reuse a cached array, this in-place subtraction would become a subtle bug. A `scaled = scaled - ...` would be equally cheap and defensively correct.

**2. `eps=0.0` with `position_handling_flag` default creates a policy where `weight_sum` can't be zero, but the `weight_sum > 0` guard is still needed**

The guard at line 395 is correct — with `eps >= 0` and `pscore > 0`, all `pi_logged` values are non-negative and `weight_sum` can only be zero if every `pi_logged` is exactly zero. This can happen at `eps=0` if the softmax concentrates all mass on a single action and no logged row chose that action. The guard is appropriate.

**3. `_validate_feedback` doesn't validate `position` range**

`bandit_log.py:560-564` validates `action` is in `[0, n_actions)` but there's no analogous check for `position` values being in a valid range. If `position` contains negative values or unexpected values, the `position == 0` mask in `run_experiment` would silently produce incorrect subsets. Consider adding:

```python
position = np.asarray(bandit_feedback["position"])
if position.min() < 0:
    raise ValueError(...)
```

This is a minor gap — the adapter doesn't need to know `n_positions`, but negative values in position would be a clear data error.

---

### Numerical Stability

**4. The softmax + SNIPW pipeline is correctly implemented.** The log-sum-exp trick (subtract max) at line 375, the `_TAU_FLOOR` guard, the `weight_sum > 0` check, and the Kish ESS formula are all standard and correct. No issues here.

**5. `weight_cv` uses `ddof=0` while `MarketingLogAdapter` uses `ddof=1`**

Documented in a comment at line 416-420, which is good. However, this inconsistency means `weight_cv` values are not comparable across adapters. If downstream code (e.g., support gates) ever applies a universal threshold to `weight_cv`, this will silently produce different behavior. Consider whether a project-wide convention should be established. Not a blocker for this PR, but worth tracking.

---

### Performance

**6. `run_experiment` allocates `(n_active, n_actions)` matrices on every call**

For the optimizer loop calling this hundreds of times on 10K rows x 34 actions, this is fine — each call creates ~5 arrays of shape `(n_active, 34)` which is ~1.3 MB total. No concern for the bundled 10K slice.

For the full ~452K-row slice (mentioned in the docstring), `marginalize` mode would allocate `(452949, 34)` arrays (~123 MB each, ~600 MB total per call). This could matter if the optimizer runs hundreds of iterations. Worth documenting in the class docstring or a comment — not a blocker but a heads-up for Issue C.

**7. `n_effective_actions` computation is vectorized and efficient.** The sort-cumsum-argmax pattern at lines 436-449 is the right approach.

---

### Type Safety

**8. Return type annotation is `dict[str, float]` but values are cast via `float()`** — This is correct and consistent with the base class.

**9. `np.ndarray` type annotations on instance variables**

The `self._action: np.ndarray` etc. annotations are fine for runtime but opaque to mypy — `np.ndarray` doesn't carry dtype/shape information. This matches the existing adapters' pattern, so no issue. Consider `npt.NDArray[np.float64]` in a future cleanup if the project adopts numpy type stubs more broadly.

---

### Test Coverage

**10. Excellent semantic test coverage.** The tests cover:
- Contract compliance (interface methods, types)
- Boundary conditions (no matching rows, zero eps, negative weights)
- Statistical sanity (uniform policy ~ logged CTR, position subsetting reduces ESS)
- Validation error paths (missing keys, non-binary reward, zero pscore)
- OBP constructor delegation (stubbed import)

**11. Missing test: `action` out-of-range in constructor validation**

`_validate_feedback` checks `action.min() < 0 or action.max() >= n_actions` but there's no unit test covering this path. Add one to `TestConstructorValidation`.

**12. Missing test: `n_rounds <= 0` or `n_actions < 2` validation**

These validation branches at lines 509-514 have no corresponding unit tests.

**13. Missing test: `action_context` with zero columns**

The validation at line 537-541 checks `action_context.shape[1] < 1` but no test exercises this path.

**14. The `from_obp` monkeypatch test is well designed** — stubbing `sys.modules` to avoid the real OBP dependency is the right pattern.

---

### Design & Architecture

**15. Clean separation between Track A (this PR) and Track B (OPE stack)**

The docstring clearly delineates scope. The SNIPW estimator here is a minimal but correct first pass. Good decision to keep it inline rather than importing from a not-yet-built OPE module.

**16. `from_obp` classmethod pattern is well-designed**

Lazy import, early validation of campaign/behavior_policy before touching OBP, and the `isinstance(bandit_feedback, dict)` guard against the timeseries-split API variant are all defensive and correct.

**17. The `bandit` extra in `pyproject.toml` is correctly structured**

Pinned to `>=0.4.1,<0.5`, included in `all`, and the pytest marker is registered. The `obp` package pulls in `torch` as a dependency (visible in `uv.lock`), which is heavy — worth noting in the install docs that `--extra bandit` adds significant weight.

---

### Code Quality

**18. Module docstring is unusually long (65 lines)**

The Sprint 34 findings are valuable documentation but the module docstring is doing double duty as both API docs and a sprint findings report. Consider moving the findings section (lines 17-55) to a separate doc or the PR description, keeping the module docstring focused on the public interface.

**19. The `_synthetic_bandit_feedback` test helper duplicates the schema contract**

If the adapter's required keys change, the test helper won't break — it'll just produce a dict that passes validation. This is fine for unit tests but worth noting: the smoke test is the real contract anchor.

**20. Integration test bypasses OBP's `pre_process` entirely**

`test_bandit_log_adapter_smoke.py:34-51` reads raw CSVs because OBP 0.4.1's preprocessor is broken on modern pandas. This is the right pragmatic choice for Sprint 35.A, but Issue B should track fixing this (either by patching OBP's call or vendoring a fixed preprocessor).

---

### Minor Nits

**21.** `bandit_log.py:339` — the f-string starts with `f"position_handling_flag must be one of "` where the leading `f` on the first string piece is unnecessary (it contains no interpolation). The second piece has the `f`. Both ruff and the runtime handle this fine; it's just cosmetic.

**22.** `bandit_log.py:210` — `np.bincount(self._action, minlength=self._n_actions)` — `self._action` is `int64` which `bincount` accepts, but the implicit assumption is that `action` values are non-negative integers. This is validated, so it's correct.

**23.** The smoke test at `test_bandit_log_adapter_smoke.py:139` uses `rankdata(data["position"].to_numpy(), "dense")` from scipy — this mirrors what OBP's own preprocessor does, but the import should be noted as a scipy dependency that's already in the core requirements.

---

### Summary

| Category | Assessment |
|----------|-----------|
| Correctness | Strong. SNIPW, Kish ESS, softmax all correct. Minor gap: no `position` range validation. |
| Numerical stability | Good. Log-sum-exp trick, tau floor, weight_sum guard all present. |
| Performance | Fine for 10K rows. Flag potential memory for 450K-row full slice. |
| Test coverage | Very thorough. A few validation branches uncovered (items 11-13). |
| Type safety | Matches existing patterns. |
| Architecture | Clean Track A/B separation. Good `from_obp` design. |
| Code quality | Well-documented. Module docstring is long but informative. |

**Verdict:** Ready to merge after addressing item 3 (position range validation) and items 11-13 (missing validation tests). Everything else is advisory.

---

datablogin:
# Claude Code Review

## Review Feedback

I now have a thorough understanding of the codebase and the PR. Here's my review:

---

## PR Review: Sprint 35.A — BanditLogAdapter

### Overall Assessment

This is a well-structured, thoroughly documented PR. The adapter follows established patterns from `MarketingLogAdapter`, the SNIPW estimator is numerically sound, and the test coverage is excellent. The code is clearly the work of someone who understood both the OBP schema and the causal-optimizer adapter contract. A few issues below, roughly ordered by severity.

---

### Bug: `_validate_feedback` shape check can pass with wrong `n_rounds`

`bandit_log.py:508-520` — The validation casts `n_rounds = int(bandit_feedback["n_rounds"])` and then checks that each 1D array has shape `(n_rounds,)`. But `n_rounds` is a user-provided scalar — if someone passes `n_rounds=500` with arrays of length 500 but an actual 600-row dataset, validation passes and the adapter stores inconsistent state. The `_action`, `_reward`, etc. arrays are derived from `np.asarray()` so their true length is authoritative; `n_rounds` is redundant metadata.

Consider adding a cross-check:

```python
actual_len = np.asarray(bandit_feedback["action"]).shape[0]
if actual_len != n_rounds:
    raise ValueError(f"n_rounds={n_rounds} but action has {actual_len} rows")
```

This is low severity (OBP always sets `n_rounds` correctly) but would catch hand-built dicts in tests.

---

### Potential Issue: `eps=0` with `tau` at the search-space floor (0.1)

`bandit_log.py:373-379` — When `eps=0` and `tau=0.1` with non-zero weights, the softmax becomes very sharp. The `scaled -= scaled.max(axis=1, keepdims=True)` log-sum-exp trick prevents overflow, but with `eps=0` the policy can concentrate all mass on a single action. This means `pi_logged` can be `~0.0` for most rows, producing importance weights that are exactly `0.0 / pscore = 0.0`. The SNIPW denominator `weight_sum` can then be very small but positive (only the few rows where the logged action matches the argmax get nonzero weight), yielding a high-variance estimate.

This isn't a *bug* — it's the correct SNIPW behavior under near-deterministic policies — but it would be worth noting in the docstring or as a log warning that ESS can drop precipitously at `eps=0`. The search space allows `eps=0.0`, so the optimizer *will* explore this corner.

---

### Minor: `n_effective_actions` in the zero-row pessimistic return is `1.0`

`bandit_log.py:363` — When no rows survive the position mask, the adapter returns `n_effective_actions: 1.0`. This is a reasonable pessimistic default, but it's semantically odd: no rows were evaluated, so the "number of effective actions" is undefined. Returning `float(self._n_actions)` (or `0.0`) might be more honest. The test at `test_bandit_log_adapter.py:301` only checks `zero_support_fraction == 1.0` but doesn't assert on `n_effective_actions` for this case, so there's no constraint from the test side.

---

### Naming inconsistency: `ess` vs `effective_sample_size`

`bandit_log.py:452` returns `"ess"` while `MarketingLogAdapter` returns `"effective_sample_size"`. If downstream code (e.g., a support gate or reporting layer) needs to compare or aggregate diagnostics across adapter types, this inconsistency will bite. If cross-adapter interoperability isn't a goal, this is fine, but worth a one-line comment explaining the divergence (the module docstring's `ddof` note covers `weight_cv` but not `ess` naming).

---

### `_validate_feedback` does double work with `np.asarray`

`bandit_log.py:516-521` — Validation calls `np.asarray()` on each array key for shape checking, and then `__init__` calls `np.asarray()` again on the same keys (lines 177-188). For a 452K-row dataset this is negligible, but on principle you could validate the shapes *after* the constructor's `np.asarray` calls rather than before. Alternatively, `_validate_feedback` could return the validated arrays to avoid the double conversion. Low priority.

---

### Integration smoke test reads CSVs by probing `obp.__file__` path internals

`test_bandit_log_adapter_smoke.py:48` — `_load_men_random_raw()` constructs a path into OBP's `site-packages` directory. This is brittle across packaging layouts (editable installs, `.egg` directories, namespace packages). The comment acknowledges the OBP 0.4.1 pandas compat issue. This is acceptable for a smoke test pinned to a specific OBP version, but should be revisited when OBP is upgraded.

---

### Test: `test_uniform_policy_has_policy_value_near_logged_ctr` relies on `eps=0.5`

`test_bandit_log_adapter.py:228-240` — The test asserts that with `eps=0.5` and zero weights, policy value matches logged CTR. But `eps=0.5` makes the policy `0.5 * uniform + 0.5 * softmax(zeros)` = `0.5 * uniform + 0.5 * uniform` = `uniform`. This is correct, but only because zero weights produce a uniform softmax. The test comment says "uniform" but doesn't explain *why* `eps=0.5` is sufficient — a reader might wonder if `eps=1.0` was intended. A clarifying note would help (or just use `eps=0.0` with zero weights, which also produces uniform via the softmax path).

---

### Good practices worth calling out

1. **Precomputation at construction** (`_item_feature_0`, `_affinity`, `_item_popularity`) — This keeps `run_experiment` cheap inside the optimizer loop. Well done.

2. **`_TAU_FLOOR` guard** — Defensive floor at `1e-6` even though the search space already forbids `tau < 0.1`. Good belt-and-suspenders for numerical stability.

3. **`_validate_feedback` as `@staticmethod`** — Clean separation; doesn't touch `self`, easy to test independently.

4. **Lazy OBP import inside `from_obp`** — Ensures the core adapter is importable without the optional extra. The `__init__.py` top-level import of `BanditLogAdapter` is safe because the module itself doesn't import `obp`.

5. **`ddof=0` for `weight_cv` with explicit cross-adapter note** — Thoughtful decision with documented rationale at `bandit_log.py:416-420`.

6. **Comprehensive validation tests** — 10 negative-path constructor tests covering every validation branch. The `from_obp` monkeypatch tests are well-crafted.

7. **Integration test separation** — `pytest.mark.slow` + `pytest.mark.obp` + `importorskip` means the fast suite never touches OBP.

---

### Type safety

The adapter's type annotations are thorough. `TYPE_CHECKING` guards for `Path` and `CausalGraph` are correct. Return types match the base class signatures. The `np.ndarray` type annotations on instance attributes don't carry dtype info (numpy typing limitation), but the code is consistent about explicit dtype casts.

One note: `_synthetic_bandit_feedback` in the unit test returns `dict[str, Any]`, which is correct — the values are heterogeneous (int, ndarray). The `Any` is unavoidable here.

---

### Summary

| Category | Verdict |
|----------|---------|
| Correctness | SNIPW, softmax, ESS all correct. No bugs found in the core estimator path. |
| Numerical stability | Good — log-sum-exp trick, tau floor, positive-pscore guard. |
| Test coverage | Excellent — 26 unit tests + 5 integration tests cover all code paths. |
| API design | Clean, follows existing adapter conventions, no OBP types leak. |
| Performance | Pre-computation pattern is appropriate for optimizer-loop usage. |
| Documentation | Module docstring is thorough (perhaps overly so for a code review, but it reads as a useful contract record). |

The `n_rounds` cross-check and the `ess` naming inconsistency are the most actionable items. The rest are minor observations. Solid work overall.

---

datablogin:
@greptileai

PR #191, opened 2026-04-20
Sprint 35 bridge: wire BanditLogAdapter into the Open Bandit OPE path

datablogin:
## Summary

Closes three narrow seams between the merged Sprint 35.A adapter ([PR #189](https://github.com/datablogin/causal-optimizer/pull/189)) and the merged Sprint 35.B OPE stack ([PR #188](https://github.com/datablogin/causal-optimizer/pull/188)) so [Issue #187](https://github.com/datablogin/causal-optimizer/issues/187) (Sprint 35.C benchmark report) can consume a ready-to-run integration path.

- **Seam 1 — position normalization.** Adds `normalize_positions(position)` in [`open_bandit.py`](causal_optimizer/benchmarks/open_bandit.py) — a dense-rank remap that produces 0-indexed contiguous integers from any integer labels. OBD loaders emit `{1, 2, 3}` (or post-filter non-contiguous labels); Track B's `_validate_positions` rejects anything else, so the bridge routes positions through this helper before the OPE wrapper consumes them.
- **Seam 2 — conditional propensity schema.** Adds `BanditLogAdapter.propensity_schema` pinned to `PROPENSITY_SCHEMA_CONDITIONAL` (Men/Random stores `action_prob` as `P(item | position) = 1 / n_items`, per the Sprint 35.A smoke-test finding). Callers thread this into `run_section_7_gates(..., schema=...)` so `propensity_sanity_gate` targets `1/n_items` instead of the default joint `1/(n_items * n_positions)`.
- **Seam 3 — OBP version provenance.** Replaces the Track B `OBP_VERSION_PLACEHOLDER` in the `run_section_7_gates` provenance dict with a runtime lookup via a new `get_obp_version()` helper. Returns installed `obp.__version__` when the `bandit` extra is present and the explicit `OBP_VERSION_UNAVAILABLE` sentinel otherwise, so provenance round-trips cleanly under graceful-degradation.

Also adds a narrow `BanditLogAdapter.to_bandit_feedback()` helper that returns an OBP-shaped dict (with normalized positions) ready for `evaluate_open_bandit_policy` and `run_section_7_gates`.

No new estimators, gates, datasets, or search-space variables. `BanditLogAdapter` and `open_bandit.py` are not rewritten.

Closes #190

## Test plan

[...]

PR #10, opened 2026-03-11
Fix known bugs: best_result, screening retry, crash guard

datablogin:
## Summary

- **Fix `ExperimentLog.best_result`**: Convert from property to method with `objective_name` and `minimize` parameters. Defaults (`"objective"`, `True`) preserve backward compatibility. Updated all callers across the codebase (engine, optimizer, tests, examples).
- **Fix screening infinite re-trigger**: Add `_screening_attempts` counter and `_max_screening_attempts=3` guard. After max retries with no important variables found, proceeds to optimization with all variables instead of looping forever.
- **Fix crashed experiments entering MAP-Elites**: Add status guard so only non-crashed experiments are added to the archive.

## Test plan

- [ ] `test_experiment_log_best_result_minimize` — verifies minimize=True returns smallest value
- [ ] `test_experiment_log_best_result_maximize` — verifies minimize=False returns largest value
- [ ] `test_experiment_log_best_result_custom_objective` — verifies custom objective_name works
- [ ] `test_experiment_log_best_result_no_kept` — verifies None returned when no kept results
- [ ] `test_experiment_log_best_result_backward_compat` — verifies default args match old behavior
- [ ] `test_screening_max_retries_limits_reattempts` — verifies screening only retries up to max
- [ ] `test_screening_max_retries_proceeds_to_optimization` — verifies forward progress after max retries
- [ ] `test_screening_increments_attempt_counter` — verifies counter increments on each attempt
- [ ] `test_crashed_experiment_not_added_to_archive` — verifies crash guard on MAP-Elites
- [ ] `test_successful_experiment_added_to_archive` — verifies non-crashed experiments still added
- [ ] All 149 existing + new tests pass
- [ ] ruff check/format clean, mypy strict clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This PR fixes three independent bugs: parameterises `ExperimentLog.best_result` with `objective_name` and `minimize` (replacing the hardcoded minimise-only property), adds a `_max_screening_attempts` guard to prevent infinite screening re-triggers, and adds a status check to keep crashed experiments out of the MAP-Elites archive. All three fixes are mechanically correct and are accompanied by dedicated unit tests.

**Key changes:**
- `causal_optimizer/types.py`: `best_result` converted from a `@property` to a proper method; fallback sentinels (`float("inf")` / `float("-inf")`) correctly respect the `minimize` flag.
[...]

PR #192, opened 2026-04-20
Sprint 35.C: first Men/Random Open Bandit benchmark report

### LINKED ISSUE #187: Sprint 35.C: run the first Men/Random benchmark and publish the report
## Context

Executes Issue C from Section 10 of the [Sprint 34 Open Bandit contract](../blob/main/thoughts/shared/docs/sprint-34-open-bandit-contract.md) and [Sprint 35 plan](../blob/main/thoughts/shared/plans/25-sprint-35-recommendation.md).

Issue C is **strictly downstream** of Issues A (adapter) and B (OPE stack + gates). Do not start this work until both A and B are merged.

## Goal

Run the first Men/Random Open Bandit benchmark and publish the report. Produce either a clean SNIPW verdict or a clean blocked-state diagnosis tied to a specific gate failure.

## Tasks

1. Run the Men/Random benchmark at 10 seeds x B20/B40/B80.
2. Run the null control under the merged Section 7 rules.
3. Publish the first Sprint 35 Open Bandit report under the same format as the Criteo report, including:
   - SNIPW-primary verdict at B80
   - DM, DR, vanilla IPW diagnostics
   - all five Section 7 gate results
   - optimizer-path provenance (Ax/BoTorch vs RF fallback)
   - OBP version pin
4. State either a clean verdict (certified / trending / not significant / near-parity) or a clean blocked-state diagnosis.

## Out of scope

- Women / All campaigns
- BTS as primary logger
- DRos, Switch-DR, MIPS, slate OPE
- claiming causal beats random before the data supports it

## Workflow

Do not start Issue C in this coordination prompt. Open it only after Issues A and B are merged and the adapter-to-OPE seam is stable.

datablogin:
## Verdict

Clean Men/Random row: `causal` and `surrogate_only` tie exactly on every seed at every budget (expected null result under the Sprint 35 null prior graph per Sprint 34 contract Section 4e), and both beat `random` at **certified significance (p = 0.0002, two-sided MWU) on every seed at every budget**. All five Sprint 34 Section 7 support gates pass cleanly on the first run. This is the "least valuable but acceptable" first-run outcome per the Sprint 34 contract Section 12 — it carries the same weight as the Criteo near-parity row.

Closes #187

## Stacking

This PR is **stacked on #191** (Sprint 35 three bridge seams). **Merge the bridge (#191) first**, then rebase/retarget this PR to `main` and merge. Do not merge to `sprint-35/open-bandit-bridge` as a final destination.

## SNIPW verdict at B80 (primary, verdict budget)

| Strategy | Mean SNIPW | Std (ddof=0) | Min | Max |
|----------|-----------:|-------------:|----:|----:|
| random | 0.005431 | 0.000221 | 0.005244 | 0.005987 |
| surrogate_only | **0.006182** | 0.000008 | 0.006158 | 0.006189 |
| causal | **0.006182** | 0.000008 | 0.006158 | 0.006189 |

`surrogate_only` and `causal` return bit-identical best-of-seed policy values on every seed (adapter returns `get_prior_graph() -> None`, so `causal` reduces to `surrogate_only`). Absolute lift over `random` at B80 = 0.000751 (~14% relative over `random`, ~20% relative over the logged-policy μ = 0.005124).

## Section 7 Support Gates (all PASS)

| Gate | Threshold | Status | Observed |
|------|-----------|--------|----------|
| 7a null control | every cell ≤ 1.05 × μ_null | **PASS** | 9/9 cells within band; max ratio = 1.0154 |
| 7b ESS floor | median ESS ≥ max(1000, n_rows / 100) = 4,530 | **PASS** | B80 median ESS = 49,867; min per-seed ESS = 4,724 |
| 7c zero-support fraction | best-of-seed ≤ 10% | **PASS** | best = 0.0 on every B80 cell |
| 7d propensity sanity | empirical mean within 10% rel of 1/n_items = 0.029412 | **PASS** | empirical = 0.029412; relative deviation ≈ 1.89e-15 |
| 7e DR/SNIPW cross-check | per-seed divergence ≤ 25% | **PASS** | max divergence = 0.48%; 0 offending seeds |

## What's in this PR

- `scripts/open_bandit_benchmark.py` — CLI runner for the Men/Random benchmark (polished to re-use `get_obp_version` and `dataset_hash` from the bridge layer rather than shadowing them).
- `scripts/_open_bandit_report_helper.py` — helper for rebuilding the report tables from the JSON artifact.
- `thoughts/shared/docs/sprint-35-open-bandit-benchmark-report.md` — full Sprint 35.C report (data provenance, configuration, per-budget and per-seed SNIPW / DM / DR tables, all five Section 7 gate details, ESS diagnostics, interpretation, scope-boundary list).

## What this PR does NOT change

- `causal_optimizer/domain_adapters/bandit_log.py` (owned by Sprint 35.A / bridge)
- `causal_optimizer/benchmarks/open_bandit.py` (owned by Sprint 35.B / bridge)
- `pyproject.toml` (bridge-owned)

## Reproducibility

Exact invocation (dataset not redistributed; unzip `open_bandit_dataset.zip` locally first):

```bash
uv run python scripts/open_bandit_benchmark.py \
  --data-path <unzipped OBD root> \
  --budgets 20,40,80 \
  --seeds 0,1,2,3,4,5,6,7,8,9 \
  --strategies random,surrogate_only,causal \
  --null-control \
  --permutation-seed 20260419 \
  --output <local_artifacts>/men_random_results.json
```

Suite wall-clock: 2,880 s (48 min) on dev laptop. Dataset load: 0.9 s. All 180 runs (90 real + 90 null) completed; no skipped seeds.

## Full report

- `thoughts/shared/docs/sprint-35-open-bandit-benchmark-report.md`

## Test plan

- [x] `uv run pytest -m "not slow" -x -q` (1374 passed, 5 skipped, 107 deselected)
- [x] `uv run ruff check .` (All checks passed!)
- [x] `uv run ruff format --check .` (180 files already formatted)
- [x] Manual re-verification of report tables against `men_random_results.json` via `scripts/_open_bandit_report_helper.py`
- [ ] Greptile review clean (5/5, 0 unresolved) — via `/gauntlet`
- [ ] `claude-review.sh` clean — via `/gauntlet`
- [ ] Human sign-off before merge

<h3>Greptile Summary</h3>

This PR delivers the Sprint 35.C Men/Random Open Bandit benchmark: a library runner (`causal_optimizer/benchmarks/open_bandit_benchmark.py`), a CLI script (`scripts/open_bandit_benchmark.py`), a report helper, unit tests, and the full benchmark report. All five Section 7 gates pass on the reported run, and all three previously flagged issues (None-poisoned null-control array, hardcoded verdict-budget 80, None SNIPW crash in the per-seed table) are resolved.

- One residual P2: when `--skip-reward-model` is passed, the DR cross-check gate is correctly marked `\"skipped\": True` in the payload but `cross_gate_passed` is set to `False`, which propagates to `all_passed=False` and the stdout gate summary — misleading for a deliberate skip.

<h3>Confidence Score: 4/5</h3>

Safe to merge; one P2 discoverability issue with the skipped DR cross-check gate should be addressed before the flag misleads future --skip-reward-model users.

All three P1/P2 findings from the prior review round are fully resolved. The single remaining issue (skipped gate driving all_passed=False) is a P2 discoverability concern, not a correctness bug in benchmark results. Scoring 4 rather than 5 to flag that the one open item is worth a small fix before merge.

scripts/open_bandit_benchmark.py — _evaluate_gates skipped-DR-cross-check path (lines 462–482).

| Filename | Overview |
|----------|----------|
| causal_optimizer/benchmarks/open_bandit_benchmark.py | New library module: OpenBanditScenario strategy loop, policy action-dist builder, reward model, and per-strategy SNIPW/DM/DR collection. Previously flagged float-cast issue in the random diagnostics path is resolved. |
| scripts/open_bandit_benchmark.py | CLI runner: loads data, orchestrates real+null-control passes, evaluates Section 7 gates, serializes artifact. Gate 7a NaN guard and float-cast fixes from prior review are applied; one remaining issue: skipped DR cross-check gate falsely drives all_passed=False. |
| scripts/_open_bandit_report_helper.py | Report helper: reads JSON artifact, prints Markdown tables. All three previously flagged issues (None-poisoned array, hardcoded verdict budget 80, None SNIPW crash on per-seed table) are now resolved. |
| tests/unit/test_open_bandit_benchmark.py | Unit tests for library module: covers strategy loop, policy dist math, reward model shape/range, null-control flag, and summary helper. |
| tests/unit/test_open_bandit_benchmark_script.py | Unit tests for CLI helpers: sanitize, parse, validate, and shared-strategy-import contract. Uses sys.path manipulation to import scripts module. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant CLI as scripts/open_bandit_benchmark.py
    participant Loader as load_men_random_slice
    participant Scenario as OpenBanditScenario
    participant Engine as ExperimentEngine / random sampler
    participant OPE as compute_snipw / dm / dr
    participant Gates as _evaluate_gates
    participant JSON as JSON artifact

    CLI->>Loader: data_path → bandit_feedback dict
    CLI->>Scenario: OpenBanditScenario(bandit_feedback)
    loop budgets × seeds × strategies (real pass)
        CLI->>Scenario: run_strategy(strategy, budget, seed)
        Scenario->>Engine: run_loop(budget) or random sample
        Engine-->>Scenario: best_params
        Scenario->>OPE: build_policy_action_dist → compute_snipw/dm/dr
        OPE-->>Scenario: OpenBanditBenchmarkResult
        Scenario-->>CLI: result
    end
    opt --null-control
        loop budgets × seeds × strategies (null pass)
            CLI->>Scenario: run_strategy(..., null_control=True, permutation_seed)
            Scenario->>OPE: permute_rewards_stratified → compute_snipw/dm/dr
            OPE-->>Scenario: OpenBanditBenchmarkResult(is_null_control=True)
            Scenario-->>CLI: result
        end
    end
    CLI->>Gates: _evaluate_gates(real_results, null_results)
    Gates-->>CLI: gate_report (all_passed, 7a–7e)
    CLI->>JSON: _sanitize_for_json → json.dump artifact
```

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%201%20code%20review%20issues.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Ascripts%2Fopen_bandit_benchmark.py%3A462-482%0A**Skipped%20DR%20cross-check%20gate%20drives%20%60all_passed%3DFalse%60**%0A%0AWhen%20%60--skip-reward-model%60%20is%20passed%20%28or%20any%20B80%20seed%20lacks%20a%20DR%20value%29%2C%20the%20else%20branch%20sets%20%60cross_gate_passed%20%3D%20False%60%2C%20which%20causes%20%60all_passed%20%3D%20all%28g%5B%22passed%22%5D%20...%29%60%20to%20be%20%60False%60.%20The%20gate%20payload%20correctly%20records%20%60%22skipped%22%3A%20True%60%2C%20but%20the%20top-level%20summary%20flag%20and%20the%20stdout%20printout%20both%20reflect%20%60passed%3DFalse%60%2C%20making%20the%20artifact%20look%20like%20a%20gate%20failure%20rather%20than%20a%20deliberate%20skip.%0A%0A%60%60%60python%0A%23%20suggested%20fix%3A%20treat%20a%20skipped%20gate%20as%20neutral%20for%20all_passed%0Across_gate_passed%20%3D%20True%20%20%20%23%20skipped%20%3D%20not%20failed%0Across_gate_payload%20%3D%20%7B%0A%20%20%20%20%22passed%22%3A%20True%2C%0A%20%20%20%20%22skipped%22%3A%20True%2C%0A%20%20%20%20%22n_b80_seeds%22%3A%20len%28b80_results%29%2C%0A%20%20%20%20%22n_dr_available%22%3A%20len%28cross_pairs%29%2C%0A%20%20%20%20%22reason%22%3A%20%22reward%20model%20disabled%20or%20DR%20unavailable%22%2C%0A%7D%0A%60%60%60%0A%0AThis%20surfaces%20as%20%60Section%207%20gates%3A%20all_passed%3DFalse%20%2F%207e%20dr_cross_check%3A%20passed%3DFalse%60%20whenever%20the%20user%20runs%20with%20%60--skip-reward-model%60%2C%20even%20though%20all%20other%20gates%20are%20clean.%0A%0A&repo=datablogin%2Fcausal-optimizer&pr=192&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=2">Fix All in Claude Code</picture></a>

`````markdown
Path: scripts/open_bandit_benchmark.py
Line: 462-482

Comment:
**Skipped DR cross-check gate drives `all_passed=False`**

When `--skip-reward-model` is passed (or any B80 seed lacks a DR value), the else branch sets `cross_gate_passed = False`, which causes `all_passed = all(g["passed"] ...)` to be `False`. The gate payload correctly records `"skipped": True`, but the top-level summary flag and the stdout printout both reflect `passed=False`, making the artifact look like a gate failure rather than a deliberate skip.

```python
# suggested fix: treat a skipped gate as neutral for all_passed
cross_gate_passed = True   # skipped = not failed
cross_gate_payload = {
    "passed": True,
    "skipped": True,
    "n_b80_seeds": len(b80_results),
    "n_dr_available": len(cross_pairs),
    "reason": "reward model disabled or DR unavailable",
}
```

This surfaces as `Section 7 gates: all_passed=False / 7e dr_cross_check: passed=False` whenever the user runs with `--skip-reward-model`, even though all other gates are clean.

`````

</details>

<sub>Reviews (5): Last reviewed commit: ["address claude review feedback (claudelo..."](https://github.com/datablogin/causal-optimizer/commit/bc11425e60565fde50ed079c87cdc95d98f42c2d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28943190)</sub>

### DIFF
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -72,11 +72,6 @@ _DEFAULT_POSITION_HANDLING_FLAG: str = "position_1_only"
 evaluation of best-of-seed policies when the optimizer picked an
 invalid literal.  Matches the Sprint 34 contract Section 4c default."""
 
-_REWARD_MODEL_N_NEIGHBORS: int = 16
-"""Nearest-neighbor count for the action-conditional reward model used
-by DM and DR.  Small enough to stay cheap on ~453K rows; large enough
-that the per-action mean is not driven by a single observation."""
-
 
 # ── Loader: raw CSV → OBP-shaped bandit_feedback ──────────────────
 
@@ -244,7 +239,14 @@ def build_policy_action_dist(
     n_rounds = adapter._n_rounds  # noqa: SLF001 — adapter is our own private surface
     n_actions = adapter._n_actions  # noqa: SLF001
 
-    # Scores for every (row, action) candidate.
+    # Scores for every (row, action) candidate. Under
+    # ``position_1_only`` the adapter instead scores the position-0
+    # subset only (see :meth:`BanditLogAdapter.run_experiment`). Here
+    # we score every row up front and then, below, overwrite the
+    # non-position-0 rows with the uniform distribution; the overwrite
+    # discards the score values on those rows, so the two paths agree
+    # element-wise on position-0 rows and the benchmark SNIPW matches
+    # the adapter's own ``policy_value``.
     item_term = w_item * adapter._item_feature_0[None, :]  # noqa: SLF001
     pop_term = w_popularity * adapter._item_popularity[None, :]  # noqa: SLF001
     affinity_term = w_affinity * adapter._affinity  # noqa: SLF001
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -486,8 +486,11 @@ class OpenBanditScenario:
             raise ValueError("permutation_seed is required when null_control=True")
 
         if null_control:
-            assert permutation_seed is not None  # narrowed by the check above
-            bf = permute_rewards_stratified(self._bandit_feedback, seed=int(permutation_seed))
+            # ``permutation_seed`` is already validated above; narrow to
+            # int without a runtime `assert` so the guard survives
+            # `python -O`.
+            perm_seed = int(permutation_seed) if permutation_seed is not None else 0
+            bf = permute_rewards_stratified(self._bandit_feedback, seed=perm_seed)
             reward_hat = compute_reward_model(bf, seed=0) if self._use_reward_model else None
         else:
             bf = self._bandit_feedback
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -270,7 +270,13 @@ def build_policy_action_dist(
             policy = policy.copy()
             policy[~mask] = 1.0 / n_actions
 
-    assert policy.shape == (n_rounds, n_actions)
+    # Shape check on computed data (not a test assertion) — use an
+    # explicit raise so the invariant survives `python -O`.
+    if policy.shape != (n_rounds, n_actions):
+        raise ValueError(
+            f"build_policy_action_dist produced shape {policy.shape}, "
+            f"expected {(n_rounds, n_actions)}"
+        )
     return np.asarray(policy, dtype=float)
 
 
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -502,7 +502,9 @@ class OpenBanditScenario:
                 if pv > best_value:
                     best_value = pv
                     best_params = params
-                    best_metrics = {k: v for k, v in metrics.items() if k != "policy_value"}
+                    # Cast to float to match the typed `diagnostics` contract
+                    # and the engine-path branch below.
+                    best_metrics = {k: float(v) for k, v in metrics.items() if k != "policy_value"}
         else:
             graph = adapter.get_prior_graph() if strategy == "causal" else None
             engine = ExperimentEngine(
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -269,6 +269,15 @@ def build_policy_action_dist(
         if not mask.all():
             policy = policy.copy()
             policy[~mask] = 1.0 / n_actions
+    elif position_flag != "marginalize":
+        # The Sprint 34 contract Section 4c only allows two values;
+        # reject unknown strings loudly rather than silently falling
+        # through to the marginalize branch. The adapter does the same
+        # validation on its own path.
+        raise ValueError(
+            f"position_handling_flag must be 'marginalize' or 'position_1_only', "
+            f"got {position_flag!r}"
+        )
 
     # Shape check on computed data (not a test assertion) — use an
     # explicit raise so the invariant survives `python -O`.
--- a/causal_optimizer/benchmarks/open_bandit_benchmark.py
+++ b/causal_optimizer/benchmarks/open_bandit_benchmark.py
@@ -1,0 +1,610 @@
+"""Sprint 35.C Open Bandit benchmark runner.
+
+Wires together:
+
+1. the ``BanditLogAdapter`` (Sprint 35.A) which exposes the six-variable
+   item-scoring search space over a logged bandit-feedback dict;
+2. the OPE stack in :mod:`causal_optimizer.benchmarks.open_bandit`
+   (Sprint 35.B) which provides SNIPW/DM/DR estimators and the
+   Section 7 support gates;
+3. the core :class:`ExperimentEngine` (for ``surrogate_only`` /
+   ``causal`` strategies) and a uniform-random sampler (for the
+   ``random`` baseline).
+
+Public entry points in this module are adapter-agnostic — they accept a
+pre-materialized OBP-style ``bandit_feedback`` dict.  The thin
+:mod:`scripts.open_bandit_benchmark` CLI translates a ``--data-path`` to
+a dict via :func:`load_men_random_slice`.
+
+Scope — Sprint 34 contract Section 3
+-----------------------------------
+
+- Men / uniform-Random campaign slice only.
+- SNIPW primary; DM and DR secondary.
+- Section 7 gate evaluation happens in the caller
+  (``scripts/open_bandit_benchmark.py``) — this module ships the
+  per-strategy loop and surfaces per-seed diagnostics.
+"""
+
+from __future__ import annotations
+
+import logging
+import math
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+from causal_optimizer.benchmarks.open_bandit import (
+    compute_dm,
+    compute_dr,
+    compute_min_propensity_clip,
+    compute_snipw,
+    permute_rewards_stratified,
+)
+from causal_optimizer.benchmarks.runner import sample_random_params
+from causal_optimizer.domain_adapters.bandit_log import BanditLogAdapter
+from causal_optimizer.engine.loop import ExperimentEngine
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+logger = logging.getLogger(__name__)
+
+
+# ── Public constants ────────────────────────────────────────────────
+
+VALID_STRATEGIES: frozenset[str] = frozenset({"random", "surrogate_only", "causal"})
+"""Strategies supported by :class:`OpenBanditScenario`.  The same three
+strategies as the Criteo and Hillstrom harnesses."""
+
+OBD_ENGINE_OBJECTIVE: str = "policy_value"
+"""Primary objective fed to :class:`ExperimentEngine`."""
+
+OBD_N_POSITIONS: int = 3
+"""Men/Random ships three positions (left, center, right) per
+Saito et al. 2021 Table 1."""
+
+_DEFAULT_POSITION_HANDLING_FLAG: str = "position_1_only"
+"""Default Section 4c position-handling flag, applied to SNIPW/DM/DR
+evaluation of best-of-seed policies when the optimizer picked an
+invalid literal.  Matches the Sprint 34 contract Section 4c default."""
+
+_REWARD_MODEL_N_NEIGHBORS: int = 16
+"""Nearest-neighbor count for the action-conditional reward model used
+by DM and DR.  Small enough to stay cheap on ~453K rows; large enough
+that the per-action mean is not driven by a single observation."""
+
+
+# ── Loader: raw CSV → OBP-shaped bandit_feedback ──────────────────
+
+
+def build_bandit_feedback_from_raw(
+    *,
+    data: pd.DataFrame,
+    item_context: pd.DataFrame,
+) -> dict[str, Any]:
+    """Translate raw Men/Random CSVs into an OBP-shaped bandit_feedback dict.
+
+    OBP 0.4.1's default ``pre_process`` calls
+    ``DataFrame.drop(..., 1)`` with a positional ``axis`` which modern
+    pandas rejects; this function bypasses that path and reads the raw
+    columns directly, matching the Sprint 35.A smoke-test approach.
+
+    Parameters
+    ----------
+    data:
+        The raw ``men.csv`` frame. Must carry ``timestamp``, ``item_id``,
+        ``position``, ``click``, ``propensity_score``, and the 34
+        ``user-item_affinity_<k>`` columns.
+    item_context:
+        The raw ``item_context.csv`` frame. Must carry ``item_id`` and
+        ``item_feature_0``.
+
+    Returns
+    -------
+    A dict with OBP keys ``n_rounds``, ``n_actions``, ``action``,
+    ``position``, ``reward``, ``pscore``, ``context``, ``action_context``.
+    Positions are re-ranked to 0-indexed contiguous integers via
+    ``scipy.stats.rankdata(..., "dense") - 1`` to match OBP's convention.
+    """
+    from scipy.stats import rankdata
+
+    required_data_cols = {"item_id", "position", "click", "propensity_score"}
+    missing = required_data_cols - set(data.columns)
+    if missing:
+        raise ValueError(
+            f"raw data frame is missing required columns: {sorted(missing)}. "
+            "Expected the Men/Random men.csv schema from "
+            "https://research.zozo.com/data_release/open_bandit_dataset.zip"
+        )
+    required_ic_cols = {"item_feature_0"}
+    if not required_ic_cols.issubset(set(item_context.columns)):
+        missing_ic = required_ic_cols - set(item_context.columns)
+        raise ValueError(f"item_context frame is missing required columns: {sorted(missing_ic)}")
+
+    if "timestamp" in data.columns:
+        data = data.sort_values("timestamp").reset_index(drop=True)
+
+    action = data["item_id"].to_numpy().astype(int)
+    position = (rankdata(data["position"].to_numpy(), "dense") - 1).astype(int)
+    reward = data["click"].to_numpy().astype(float)
+    pscore = data["propensity_score"].to_numpy().astype(float)
+
+    affinity_cols = sorted(
+        (c for c in data.columns if c.startswith("user-item_affinity_")),
+        key=lambda c: int(c.rsplit("_", 1)[-1]),
+    )
+    context = data[affinity_cols].to_numpy().astype(float)
+
+    # Sort item_context by item_id so column 0 of ``action_context``
+    # matches action index 0 (BanditLogAdapter indexes into this array
+    # by raw item id).
+    if "item_id" in item_context.columns:
+        item_context = item_context.sort_values("item_id")
+    action_context = item_context[["item_feature_0"]].to_numpy().astype(float)
+
+    # Defensive sanity: action indices must lie in [0, n_actions).
+    n_actions = int(action_context.shape[0])
+    if int(action.min()) < 0 or int(action.max()) >= n_actions:
+        raise ValueError(
+            f"raw action ids range [{int(action.min())}, {int(action.max())}] but "
+            f"item_context provides {n_actions} items. The loader cannot align the "
+            "action column to the item_context rows."
+        )
+
+    # Widen context to at least n_actions columns so the adapter's
+    # per-candidate affinity lookup (context[:, :n_actions]) remains well
+    # defined even under small fixtures. The Men/Random slice already
+    # has 34 affinity columns, so this pad is a no-op on real data.
+    if context.shape[1] < n_actions:
+        pad = np.zeros((context.shape[0], n_actions - context.shape[1]), dtype=float)
+        context = np.concatenate([context, pad], axis=1)
+
+    return {
+        "n_rounds": int(len(data)),
+        "n_actions": n_actions,
+        "action": action,
+        "position": position,
+        "reward": reward,
+        "pscore": pscore,
+        "context": context,
+        "action_context": action_context,
+    }
+
+
+def load_men_random_slice(*, data_path: Path) -> dict[str, Any]:
+    """Load the full Men/Random slice from a local Open Bandit Dataset root.
+
+    Expects ``data_path`` to point at a directory laid out like::
+
+        data_path/
+          random/
+            men/
+              men.csv
+              item_context.csv
+
+    This matches the layout inside
+    ``open_bandit_dataset.zip`` from
+    ``https://research.zozo.com/data_release/``.
+
+    Raises
+    ------
+    FileNotFoundError:
+        When either CSV is missing under the expected layout.
+    """
+    import pandas as pd
+
+    root = Path(data_path)
+    men_csv = root / "random" / "men" / "men.csv"
+    item_csv = root / "random" / "men" / "item_context.csv"
+    if not men_csv.is_file():
+        raise FileNotFoundError(
+            f"Men/Random men.csv not found at {men_csv}. Expected the "
+            "open_bandit_dataset.zip layout with a ``random/men/`` "
+            "subdirectory under --data-path."
+        )
+    if not item_csv.is_file():
+        raise FileNotFoundError(f"Men/Random item_context.csv not found at {item_csv}.")
+    data = pd.read_csv(men_csv, index_col=0)
+    item_context = pd.read_csv(item_csv, index_col=0)
+    return build_bandit_feedback_from_raw(data=data, item_context=item_context)
+
+
+# ── Policy action_dist builder ────────────────────────────────────
+
+
+def build_policy_action_dist(
+    *, adapter: BanditLogAdapter, parameters: dict[str, Any]
+) -> np.ndarray:
+    """Return the ``[n_rounds, n_actions]`` policy distribution for ``parameters``.
+
+    Mirrors the adapter's softmax-over-linear-scores + epsilon-uniform
+    math in :meth:`BanditLogAdapter.run_experiment`. The two paths share
+    an identical closed-form definition of the policy, so feeding the
+    dist returned here into :func:`compute_snipw` reproduces the adapter's
+    own ``policy_value`` (up to numerical noise from the pscore clip).
+
+    For ``position_handling_flag == "position_1_only"``, rows that fall
+    outside position 0 have their row set to the uniform distribution,
+    which yields a neutral SNIPW contribution once the caller subsets to
+    position 0 for the verdict. Callers that want strict subsetting can
+    filter the bandit_feedback and the action_dist by
+    ``position == 0`` in lockstep.
+    """
+    tau = float(parameters.get("tau", 1.0))
+    eps = float(parameters.get("eps", 0.0))
+    w_item = float(parameters.get("w_item_feature_0", 0.0))
+    w_affinity = float(parameters.get("w_user_item_affinity", 0.0))
+    w_popularity = float(parameters.get("w_item_popularity", 0.0))
+    position_flag = str(parameters.get("position_handling_flag", _DEFAULT_POSITION_HANDLING_FLAG))
+
+    n_rounds = adapter._n_rounds  # noqa: SLF001 — adapter is our own private surface
+    n_actions = adapter._n_actions  # noqa: SLF001
+
+    # Scores for every (row, action) candidate.
+    item_term = w_item * adapter._item_feature_0[None, :]  # noqa: SLF001
+    pop_term = w_popularity * adapter._item_popularity[None, :]  # noqa: SLF001
+    affinity_term = w_affinity * adapter._affinity  # noqa: SLF001
+    scores = item_term + pop_term + affinity_term  # shape (n_rounds, n_actions)
+
+    safe_tau = max(tau, 1e-6)
+    scaled = scores / safe_tau
+    scaled = scaled - scaled.max(axis=1, keepdims=True)  # numerical stability
+    exp_scores = np.exp(scaled)
+    softmax = exp_scores / exp_scores.sum(axis=1, keepdims=True)
+    uniform = np.full_like(softmax, 1.0 / n_actions)
+    policy = (1.0 - eps) * softmax + eps * uniform
+
+    if position_flag == "position_1_only":
+        # Rows outside position 0 get the uniform distribution so the
+        # row-sums remain valid probabilities and SNIPW on the full
+        # feedback degrades to a uniform-policy contribution on the
+        # masked rows. The per-verdict quote subsets to position 0.
+        mask = adapter._position == 0  # noqa: SLF001
+        if not mask.all():
+            policy = policy.copy()
+            policy[~mask] = 1.0 / n_actions
+
+    assert policy.shape == (n_rounds, n_actions)
+    return np.asarray(policy, dtype=float)
+
+
+# ── Reward model for DM / DR ──────────────────────────────────────
+
+
+def compute_reward_model(bandit_feedback: dict[str, Any], *, seed: int) -> np.ndarray:
+    """Return an ``[n_rounds, n_actions]`` reward-model estimate.
+
+    Used by DM and DR as the ``reward_hat`` surface.  The estimator is
+    deliberately simple: for each action ``a`` we estimate
+    ``E[reward | action=a]`` as the mean of the logged rewards for rows
+    that selected that action, and broadcast that scalar per row.  This
+    keeps the reward model honest enough to catch blatant DM bias while
+    staying well under the per-budget runtime budget.
+
+    ``seed`` is accepted for API consistency; the estimator is
+    deterministic given ``bandit_feedback``.
+
+    On Men/Random the per-action empirical CTR ranges from roughly
+    ``0.003`` to ``0.010`` (overall CTR ~0.005), so the scalar-per-action
+    reward hat is a reasonable zero-context baseline.
+    """
+    del seed  # reserved
+    action = np.asarray(bandit_feedback["action"], dtype=int)
+    reward = np.asarray(bandit_feedback["reward"], dtype=float)
+    n_actions = int(bandit_feedback["n_actions"])
+    n_rounds = int(bandit_feedback["n_rounds"])
+
+    # Per-action mean reward; fall back to the global mean for actions
+    # that never appear (should not happen on Men/Random but keeps the
+    # estimator well defined on edge cases).
+    action_sum = np.bincount(action, weights=reward, minlength=n_actions).astype(float)
+    action_count = np.bincount(action, minlength=n_actions).astype(float)
+    global_mean = float(reward.mean()) if reward.size > 0 else 0.0
+    with np.errstate(invalid="ignore", divide="ignore"):
+        safe_counts = np.maximum(action_count, 1)
+        action_mean = np.where(action_count > 0, action_sum / safe_counts, global_mean)
+    # Clamp to [0, 1] — per-action means are already in that range on
+    # binary rewards, the clamp is a defence against rare numerical drift.
+    action_mean = np.clip(action_mean, 0.0, 1.0)
+
+    # Broadcast to (n_rounds, n_actions): every row sees the same
+    # per-action scalar. Richer reward models are deferred to Sprint 36+.
+    reward_hat = np.broadcast_to(action_mean[None, :], (n_rounds, n_actions)).copy()
+    return np.asarray(reward_hat, dtype=float)
+
+
+# ── Result dataclass ──────────────────────────────────────────────
+
+
+@dataclass
+class OpenBanditBenchmarkResult:
+    """Result of running one strategy on the Open Bandit benchmark.
+
+    Attributes
+    ----------
+    strategy:
+        One of ``"random"`` / ``"surrogate_only"`` / ``"causal"``.
+    budget:
+        Number of experiments evaluated by the strategy.
+    seed:
+        RNG seed for the strategy's optimizer.
+    is_null_control:
+        ``True`` when the reward column was permuted before evaluation.
+    permutation_seed:
+        The permutation seed used for the null control, or ``None``.
+    policy_value_snipw:
+        Primary verdict estimator. Reported on the best-of-budget
+        parameters for every strategy.
+    policy_value_dm:
+        Secondary DM estimate, ``None`` when the reward model is disabled.
+    policy_value_dr:
+        Secondary DR estimate, ``None`` when the reward model is disabled.
+    selected_parameters:
+        Best-of-budget parameter dict returned by the strategy.
+    runtime_seconds:
+        Wall-clock runtime of the strategy run.
+    diagnostics:
+        The Section 4d diagnostic dict
+        (``ess``, ``weight_cv``, ``max_weight``,
+        ``zero_support_fraction``, ``n_effective_actions``) plus
+        ``n_clipped_rows``.
+    """
+
+    strategy: str
+    budget: int
+    seed: int
+    is_null_control: bool
+    permutation_seed: int | None
+    policy_value_snipw: float
+    policy_value_dm: float | None
+    policy_value_dr: float | None
+    selected_parameters: dict[str, Any] | None = None
+    runtime_seconds: float = 0.0
+    diagnostics: dict[str, float] = field(default_factory=dict)
+
+
+# ── Strategy scenario ─────────────────────────────────────────────
+
+
+class OpenBanditScenario:
+    """Top-level Open Bandit benchmark scenario.
+
+    Wraps a pre-materialized OBP-shaped bandit-feedback dict in a
+    :class:`BanditLogAdapter` and runs one strategy at a given
+    ``(budget, seed)`` pair.
+
+    Parameters
+    ----------
+    bandit_feedback:
+        OBP-shaped dict. Not mutated.
+    use_reward_model:
+        When ``True`` (default), compute a per-action reward-model
+        baseline and report DM and DR estimates on the best-of-budget
+        policy. Set to ``False`` to skip reward-model fitting.
+    min_propensity_clip:
+        Section 5c floor for SNIPW/DR. Defaults to
+        ``1 / (2 * n_actions * n_positions)`` per contract Section 5c.
+    """
+
+    def __init__(
+        self,
+        *,
+        bandit_feedback: dict[str, Any],
+        use_reward_model: bool = True,
+        min_propensity_clip: float | None = None,
+    ) -> None:
+        self._bandit_feedback = bandit_feedback
+        self._use_reward_model = use_reward_model
+
+        n_actions = int(bandit_feedback["n_actions"])
+        # Infer n_positions from the data; callers should already have a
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
@greptileai

**Comment by datablogin** [comment]
# Claude Code Review

## Review Feedback

I now have a thorough understanding of the code. Here's my review:

---

## PR Review: Sprint 35.C Open Bandit Benchmark

Overall this is a well-structured, well-documented PR. The benchmark harness, CLI, tests, and report are all cohesive. Below are the issues I found, organized by severity.

### Bugs / Correctness Issues

**1. `build_policy_action_dist` affinity term doesn't match adapter scoring (Medium)**

`open_bandit_benchmark.py:250` computes:
```python
affinity_term = w_affinity * adapter._affinity  # all rows
```

But the adapter at `bandit_log.py:380` computes:
```python
affinity_term = w_affinity * self._affinity[mask]  # position-masked rows only
```

When `position_handling_flag == "position_1_only"`, the adapter subsets rows *before* computing scores. `build_policy_action_dist` computes scores for *all* rows and then overwrites non-position-0 rows with uniform. The scores at position-0 rows match (affinity for those rows is the same), so the final SNIPW for the position-0 subset agrees — but this is fragile and only works because the overwrite to uniform on non-position-0 rows discards the scores anyway. Worth a comment noting this explicitly, since a future maintainer might assume the two paths are literally identical element-wise.

**2. `_REWARD_MODEL_N_NEIGHBORS` is defined but never used (Low)**

`open_bandit_benchmark.py:75` defines `_REWARD_MODEL_N_NEIGHBORS = 16` with a docstring about "nearest-neighbor count for the action-conditional reward model." But `compute_reward_model` uses a simple per-action mean, not k-NN. The constant is dead code — either the docstring is stale from an earlier design or the implementation was simplified. Should be removed to avoid confusion.

**3. `_sanitize_for_json` doesn't handle dict keys that are non-string (Low)**

`scripts/open_bandit_benchmark.py:84` sanitizes values but not keys. The code passes `dataclasses.asdict()` results which always have string keys, so this isn't currently a problem, but it's worth noting for robustness. Not a blocker.

### Design / Code Quality

**4. Heavy reliance on private adapter attributes (Medium)**

`build_policy_action_dist` accesses 6 private attributes (`_n_rounds`, `_n_actions`, `_item_feature_0`, `_item_popularity`, `_affinity`, `_position`) from `BanditLogAdapter`. The `# noqa: SLF001` comments acknowledge this. Consider whether the adapter should expose a `build_action_dist(parameters)` method that encapsulates this logic, keeping the softmax math in one place. This would eliminate the dual-maintenance burden (adapter scoring vs. benchmark scoring) and remove the `SLF001` suppressions. Not blocking this PR, but this should be addressed soon.

**5. Test script uses `sys.path` manipulation**

`test_open_bandit_benchmark_script.py:15-18` inserts the scripts directory into `sys.path` to import the CLI module. This works but is brittle. A cleaner approach would be a small `__init__.py` or to move the testable logic into the package and keep the script as a thin wrapper. Minor — common pattern in this codebase.

**6. `_print_summary` sort order may be confusing**

`scripts/open_bandit_benchmark.py:203` sorts by `(is_null, strategy, budget)`, which groups null=False before null=True (since `False < True`), then alphabetically by strategy. This means `causal` appears before `random` before `surrogate_only`, not the natural `random → surrogate_only → causal` progression. Minor UX issue.

### Numerical / Statistical

**7. Softmax numerical stability is correct**

Both the adapter and `build_policy_action_dist` correctly subtract the row-max before `np.exp()`. The `max(tau, 1e-6)` floor prevents division by zero. The epsilon-mixing ensures no action probability is exactly zero. All good.

**8. `compute_reward_model` is deliberately simple — acknowledged**

The per-action empirical mean reward model is adequate for the DM/DR cross-check gate. The docstring and report correctly note this is a zero-context baseline. The `np.errstate` guard and `np.where` fallback for never-seen actions is correct. The `[0, 1]` clamp is appropriate for binary rewards.

**9. `np.std(ddof=0)` is intentional**

Population std rather than sample std is used for summary tables. This matches the report tables and is consistent across the codebase. Correct for reporting population statistics over a fixed set of seeds.

### Test Coverage

**10. Tests cover the critical paths well but have gaps:**

- `build_bandit_feedback_from_raw` validation errors (missing columns) are not tested — only the happy path and adapter acceptance.
- No test for `budget <= 0` raising `ValueError` (the CLI validator test covers `sys.exit` but not the scenario-level raise).
- No test for `null_control=True` without `permutation_seed` raising `ValueError`.
- No test that `use_reward_model=False` produces `None` for DM/DR.
- The report helper (`_open_bandit_report_helper.py`) has zero test coverage. Even one test exercising `main()` with a synthetic JSON artifact would catch regressions.
- `_evaluate_gates` in the CLI script has no direct unit test. The gate logic is complex enough to warrant tests for edge cases (empty `null_results`, missing ESS diagnostics, partial DR availability).

**11. `test_position_1_only_mask_restricts_rows` tests the uniform rows but not the position-0 rows**

The test checks that non-position-0 rows sum to 1, but doesn't verify that position-0 rows have the *expected non-uniform* distribution. Checking that at least some position-0 rows have non-uniform mass would strengthen this test.

### Performance

**12. Reward model recomputed per null-control call**

In `run_strategy`, when `null_control=True`, `compute_reward_model(bf, seed=0)` is called fresh on each call. For 90 null-control runs this means 90 per-action mean computations on the full ~453K row dataset. Each is O(n) and fast, but if the same permutation seed is used across seeds (which it is — one fixed permutation seed), the permuted `bf` is identical across all `(strategy, seed)` calls at the same permutation seed, so the reward model is recomputed identically 90 times. Could be cached, but the cost is negligible (~ms each). Not blocking.

**13. `build_policy_action_dist` allocates full `(n_rounds, n_actions)` arrays**

For 452K rows × 34 actions this is about 120MB per call (float64). Three such arrays are allocated simultaneously (item_term, pop_term, affinity_term + scores + scaled + exp_scores + softmax + policy). Peak memory is ~700MB. Acceptable for a benchmark script, but worth noting.

### Report Quality

**14. Report is thorough and well-structured.** The provenance section, gate details, per-seed tables, and scope boundaries are exemplary. The interpretation section correctly explains the causal-vs-surrogate tie as a consequence of the null prior graph.

### Summary

| Severity | Count | Key items |
|----------|-------|-----------|
| Medium | 2 | Dead constant `_REWARD_MODEL_N_NEIGHBORS`; private attribute coupling |
| Low | 3 | Affinity term comment gap; sort order; sanitizer edge case |
| Test gaps | 5 | Missing error-path tests; no report-helper tests; no gate-evaluation tests |

**Recommendation:** Merge-ready after removing the dead `_REWARD_MODEL_N_NEIGHBORS` constant and adding a brief comment in `build_policy_action_dist` explaining why the all-rows-then-overwrite approach is equivalent to the adapter's subset-then-score approach. The test gaps are real but not blocking — they can be addressed in a follow-up.

---

datablogin:
# Claude Code Review

## Review Feedback



Here is my review of the PR.

---

## PR Review: Sprint 35.C Open Bandit Benchmark

Overall this is well-structured, well-documented code with solid test coverage. The interface integrations all check out — every function call matches its dependency's signature correctly. Below are the issues I found, organized by severity.

---

### Bugs / Correctness Issues

**1. `_evaluate_gates` mixes all strategies into ESS/zero-support gates (medium severity)**

`scripts/open_bandit_benchmark.py:416` — `b80_results` collects **all** strategies at the verdict budget, then feeds all their ESS/zero-support values into a single gate call. This means `random`'s high-ESS values inflate the median ESS and mask any potential issue with `surrogate_only`/`causal`. The gate docstring says "per-seed diagnostics captured on the B80 real results" but the semantics of the gates from `open_bandit.py` seem per-strategy. If the intent is to gate each strategy independently, this is a bug. If aggregating across strategies is intentional, a comment explaining why would help.

**2. `build_policy_action_dist` uses `assert` for a runtime invariant (low severity)**

`open_bandit_benchmark.py:273` — `assert policy.shape == (n_rounds, n_actions)` will be stripped under `python -O`. Since this is a shape check on computed data (not a test assertion), use an explicit `if` + `raise ValueError` instead, or accept that this is debug-only.

**3. Null-control reward model recomputed on every `run_strategy` call (performance)**

`open_bandit_benchmark.py:485` — When `null_control=True`, `compute_reward_model(bf, seed=0)` is called on every `(strategy, budget, seed)` invocation. Since the permuted feedback is deterministic for a given `permutation_seed`, the reward model is identical across all 30 null-control cells. At 452K rows this is cheap but wasteful — the scenario could cache it keyed on `permutation_seed`.

---

### Design / Architecture Concerns

**4. Heavy reliance on private adapter attributes (`# noqa: SLF001`)**

`open_bandit_benchmark.py:239-252,268` — `build_policy_action_dist` accesses six private attributes (`_n_rounds`, `_n_actions`, `_item_feature_0`, `_item_popularity`, `_affinity`, `_position`). The `# noqa: SLF001` comments acknowledge this, but it creates a tight coupling where any internal refactor of `BanditLogAdapter` silently breaks this module. Consider adding a thin public accessor to the adapter (e.g. `adapter.scoring_components() -> ScoringComponents`) or at minimum adding an integration test that asserts these attributes exist, so breakage surfaces immediately.

**5. `_sanitize_for_json` doesn't handle `set` or `frozenset` types**

`scripts/open_bandit_benchmark.py:81-98` — If any dataclass field or diagnostic dict ever contains a `set` (e.g., `VALID_STRATEGIES` is a `frozenset`), this will pass through to `json.dump` and crash. Not currently triggered but fragile if the output schema evolves.

**6. Script test imports via `sys.path` manipulation**

`test_open_bandit_benchmark_script.py:15-18` — Inserting the `scripts/` directory into `sys.path` is brittle and can shadow installed packages. Consider making `scripts/open_bandit_benchmark.py` a proper entry point in `pyproject.toml` or using `importlib.util.spec_from_file_location` for the test.

---

### Test Coverage Gaps

**7. No test for `_evaluate_gates`**

The gate evaluation logic in `scripts/open_bandit_benchmark.py:388-524` is ~140 lines of non-trivial logic (null-control band check, ESS aggregation, DR cross-check wiring, edge cases for missing DR) with no dedicated unit test. The null-control band arithmetic (`mu_null`, `threshold`, `finite_null_values` filtering) is particularly worth testing since it determines PASS/FAIL verdicts.

**8. No test for negative budget**

`open_bandit_benchmark.py:477-478` validates `budget <= 0` but no test exercises `budget < 0` (only `budget=0` is indirectly covered by the script-level `_validate_budgets` test).

**9. No test for `use_reward_model=False` path**

The `OpenBanditScenario(use_reward_model=False)` path is documented and accepted by the constructor but never tested. A test should verify that `policy_value_dm` and `policy_value_dr` are both `None` in the result.

**10. `test_position_1_only_mask_restricts_rows` doesn't verify the *values* of masked rows**

`test_open_bandit_benchmark.py:272-274` — The test checks that off-position rows sum to 1 but doesn't verify they are *uniform* (i.e., each element equals `1/n_actions`). The sum-to-1 check is necessary but not sufficient.

---

### Numerical / Statistical Concerns

**11. Softmax numerical stability is correct but `tau` floor could be tighter**

`open_bandit_benchmark.py:255` — `safe_tau = max(tau, 1e-6)` is fine, but the search space lower bound for `tau` is 0.1, so the 1e-6 floor should never activate on valid optimizer output. Worth noting in a comment that this guard is only for malformed external callers.

**12. `compute_reward_model` is deterministic despite accepting a `seed` parameter**

`open_bandit_benchmark.py:280-319` — The `seed` parameter is immediately `del`-eted. The docstring explains it's "reserved for API consistency," which is reasonable, but from a caller's perspective this is a surprising no-op. Consider making it `_seed` or `seed: int = 0` with a default to signal it's unused.

---

### Style / Minor

**13. Unused import in `scripts/open_bandit_benchmark.py`**

Line 65: `from causal_optimizer.domain_adapters.bandit_log import BanditLogAdapter` — `BanditLogAdapter` is listed in `__all__` (line 531) as a re-export but is never used directly in the script body. If it's only for re-export convenience, that's fine but worth a comment.

**14. `_open_bandit_report_helper.py` has no tests**

This is a report-drafting-time helper, so it may be intentionally untested, but the MWU classification logic (`_classify`) and the null-control table math duplicate logic from `_evaluate_gates` and could drift.

**15. Duplicate DataFrame construction in `TestBuildFromRaw`**

`test_open_bandit_benchmark.py:90-156` — Three test methods construct nearly identical DataFrames. Extract a shared fixture (or use the existing `_synthetic_feedback` pattern) to reduce duplication and maintenance burden.

---

### What's Done Well

- Excellent docstrings throughout — the module-level docstring, per-function parameter docs, and inline comments are thorough and reference the Sprint 34 contract sections.
- Clean separation between the benchmark runner module (library) and the CLI script.
- `_sanitize_for_json` handles the full numpy type zoo correctly (integer, floating, bool_, ndarray).
- The `OpenBanditBenchmarkResult` dataclass is well-designed with sensible defaults.
- The SNIPW re-evaluation after best-of-budget selection (lines 537-550) is the correct approach — it ensures the reported value uses the contract's propensity clip rather than the adapter's internal unclipped SNIPW.
- The report is impressively detailed with full provenance, reproducibility command, and honest scope boundaries.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of all the code. Let me compile my review.

---

## PR Review: Sprint 35.C Open Bandit Benchmark

Overall this is a well-structured, well-documented PR. The benchmark runner, CLI script, and tests are thorough. Here are my findings organized by severity.

### Bugs / Correctness Issues

**1. `build_policy_action_dist` accesses adapter internals — fragile coupling and potential semantic mismatch**
`open_bandit_benchmark.py:239-252`

The function accesses 6 private attributes of `BanditLogAdapter` via `adapter._n_rounds`, `adapter._item_feature_0`, etc. with `# noqa: SLF001` suppressions. This duplicates the softmax + epsilon logic that already lives inside `BanditLogAdapter.run_experiment()`. If the adapter's scoring logic ever changes (new terms, different normalization), this copy will silently diverge. Consider adding a public method to `BanditLogAdapter` like `build_action_dist(parameters) -> np.ndarray` that both `run_experiment()` and this benchmark can call, eliminating the duplication and the private-access surface.

**2. `_evaluate_gates` uses `max(budgets)` as verdict budget — silent wrong answer if budgets don't include 80**
`scripts/open_bandit_benchmark.py:404`

```python
verdict_budget = max(budgets)
```

If a user passes `--budgets 20,40`, the verdict budget becomes 40, but the report and contract assume B80 is the verdict budget. The ESS floor, cross-check, and zero-support gates all run on whatever `max(budgets)` is without any warning. Consider either validating that 80 is in `budgets` or at least logging a warning when the verdict budget differs from the expected 80.

**3. `assert` used for control flow in production code**
`open_bandit_benchmark.py:489`

```python
assert permutation_seed is not None  # narrowed by the check above
```

This assert is removed by `python -O`. The guard on line 485-486 correctly raises `ValueError`, so the assert is a type-narrowing hint only. Replace with a proper guard or use `typing.assert_never` / `cast` patterns to narrow for mypy without runtime `assert`.

### Design / API Issues

**4. Script imported via `sys.path` manipulation in tests**
`test_open_bandit_benchmark_script.py:15-19`

```python
_SCRIPTS_DIR = str(Path(__file__).resolve().parent.parent.parent / "scripts")
if _SCRIPTS_DIR not in sys.path:
    sys.path.insert(0, _SCRIPTS_DIR)
import open_bandit_benchmark as obb
```

This `sys.path` mutation is fragile and can shadow installed packages. If `scripts/` ever has an `open_bandit_benchmark.py` that conflicts with the library module, imports break silently. Consider making `scripts/` a package, or moving the testable CLI logic (parsing, sanitization, validation) into the library module and testing it there directly.

**5. `_sanitize_for_json` doesn't handle `set` or `frozenset`**
`scripts/open_bandit_benchmark.py:81-98`

If any gate result dataclass ever includes a `set` (e.g., `offending_seeds` from the DR cross-check gate), the sanitizer will pass it through to `json.dump`, which will fail. Not a current bug, but the pattern is brittle. Consider adding a `set`/`frozenset` case that converts to `list`.

**6. Bare `except Exception` swallows errors silently**
`scripts/open_bandit_benchmark.py:275-282, 310-317`

The main loop catches any `Exception` and logs a warning, then continues. This is reasonable for a long benchmark run, but it also swallows things like `KeyError`, `TypeError`, or `MemoryError` that indicate genuine code bugs rather than transient failures. Consider catching a narrower exception type, or at minimum re-raising after a count threshold (e.g., if >50% of cells fail, abort).

### Numerical / Statistical Issues

**7. `compute_reward_model` ignores context — acceptable now, but DM estimates are near-constant**
`open_bandit_benchmark.py:286-325`

The reward model is a per-action mean with no context features. The report correctly notes that DM is "almost policy-agnostic" and only a sanity check. This is fine for Sprint 35, but worth noting that the DR cross-check gate (7e) may not catch SNIPW variance issues because DR with a zero-context reward model is nearly identical to SNIPW. The gate passes cleanly here but provides less independent validation than a context-aware model would.

**8. Position handling in `build_policy_action_dist` vs. SNIPW evaluation**
`open_bandit_benchmark.py:263-271`

For `position_1_only`, non-position-0 rows get the uniform distribution. When SNIPW is then computed over the full `bandit_feedback`, these rows contribute `(1/n_actions) / pscore * reward` terms. The comment says this is a "neutral SNIPW contribution," which is only true if the uniform distribution matches the logging policy exactly (which it does for Men/Random's conditional 1/n_actions propensity). On a different dataset with non-uniform propensities, this would introduce bias. The coupling to the conditional propensity schema should be documented more explicitly.

### Test Coverage Gaps

**9. No test for `use_reward_model=False` path**

There's a test for `use_reward_model=True` (`test_dr_estimate_present_when_reward_model_available`) but no corresponding test asserting that `policy_value_dm is None` and `policy_value_dr is None` when `use_reward_model=False`.

**10. No test for `build_bandit_feedback_from_raw` error paths**

The function has 3 validation checks (missing data columns, missing item_context columns, action-id range mismatch) but none of these raise-on-error paths are tested. The `positions_are_zero_indexed_contiguous` test is good but doesn't cover the error cases.

**11. No test for the `_evaluate_gates` function**

The gates evaluation logic in the CLI script is untested. The individual gate functions (`ess_gate`, etc.) are presumably tested in their own module, but the wiring in `_evaluate_gates` — particularly the null-control gate, the cross-check skip logic, and the `all_passed` aggregation — is covered only by the end-to-end benchmark run on real data.

**12. No test for `summarize_strategy_budget` filtering null-control results**

The summary function filters out `is_null_control=True` results, but this behavior isn't tested. Adding a null-control result to the input and asserting it's excluded from the summary would be a useful edge-case test.

### Minor / Style Issues

**13. Unused import in `__all__`**
`scripts/open_bandit_benchmark.py:541-542`

```python
__all__ = [
    "BanditLogAdapter",
    "build_policy_action_dist",
    ...
]
```

`BanditLogAdapter` and `build_policy_action_dist` are re-exported from the script's `__all__` even though the script is never imported as a library (it's a CLI entry point). The comment says "re-exports so the CLI surface references land-of-truth imports" but this `__all__` on a `__main__` script is dead code.

**14. `_open_bandit_report_helper.py` hardcodes strategy order**
`_open_bandit_report_helper.py:186`

```python
strategies = ["random", "surrogate_only", "causal"]
```

This list is duplicated from `VALID_STRATEGIES` rather than derived from the artifact data. If a new strategy is added to the benchmark, the report helper will silently skip it in the per-seed table.

**15. Report helper has a double-nested `payload.get("gates", {}).get("gates", {})` access**
`_open_bandit_report_helper.py:148`

```python
gates = payload.get("gates", {}).get("gates", {})
```

The artifact structure is `{"gates": {"all_passed": ..., "verdict_budget": ..., "gates": {...}}}`. This double-nested `get` is correct but fragile — it would silently return `{}` if the artifact schema changes, producing an empty gate report section with no error.

### Performance

**16. `permute_rewards_stratified` is called per (strategy, budget, seed) in null control**
`open_bandit_benchmark.py:490`

With the same permutation seed, the permutation output is deterministic. Yet `run_strategy` calls it once per null-control cell. On 452K rows × 90 null cells, this is 90 redundant shuffles. The scenario could cache the permuted feedback per `permutation_seed` to avoid this.

### Summary

This is a solid first benchmark implementation. The core numerical logic (SNIPW/DM/DR, softmax policy, propensity clipping) is correct and well-commented. The main areas for improvement are:

1. **Eliminate the duplicated softmax logic** between `build_policy_action_dist` and `BanditLogAdapter.run_experiment` — this is the highest-risk item for future drift.
2. **Add tests for error paths** in the loader and for the `_evaluate_gates` wiring.
3. **Cache the permuted feedback** to avoid redundant work in null-control runs.
4. **Replace `assert` with proper guard** for the permutation_seed narrowing.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of all the code. Let me write the review.

---

## PR Review: Sprint 35.C Open Bandit Benchmark

Overall this is a well-structured, thoroughly documented PR. The benchmark runner, CLI script, tests, and report are all coherent and carefully built. Below are my findings organized by severity.

---

### Bugs / Correctness Issues

**1. `build_policy_action_dist` diverges from adapter on `position_1_only` (Medium)**

`open_bandit_benchmark.py:250-253` — The benchmark's `build_policy_action_dist` computes `affinity_term = w_affinity * adapter._affinity` over **all** rows, then masks non-position-0 rows to uniform afterward. The adapter's `run_experiment` (`bandit_log.py:380`) computes `affinity_term = w_affinity * self._affinity[mask]` — it only indexes the position-0 rows when building scores.

For rows **at** position 0, the two paths agree because both use the same affinity values. But the intermediate computation allocates a full `(n_rounds, n_actions)` scores array even though the non-position-0 rows are overwritten. This is wasteful on 452K rows but not incorrect, because the overwrite discards those values. The comment on line 242-249 explains this correctly. However, the code relies on the position-0 affinity slices being identical — if `adapter._affinity` were somehow row-order-dependent on the mask, the paths would diverge. This looks safe as-is, but it's fragile coupling to adapter internals.

**2. `_evaluate_gates` DR/SNIPW alignment issue (Low-Medium)**

`scripts/open_bandit_benchmark.py:455-457` — The DR/SNIPW cross-check gate filters `dr_per_seed` with `if r.policy_value_dr is not None` but takes `snipw_per_seed` unfiltered. Then it checks `len(dr_per_seed) == len(snipw_per_seed)`. If even one B80 result has DR = None while all others have DR, the lists won't match and the gate silently falls through to `cross_gate_passed = False` with a "reward model disabled" reason. That's misleading — the reward model wasn't disabled; one seed just failed. Consider filtering both lists in lockstep by seed to avoid this.

**3. Null-control uses `null_results[0].permutation_seed` without checking emptiness (Low)**

`scripts/open_bandit_benchmark.py:507` — The `null_results` block is guarded by `if null_results:`, so `null_results[0]` is safe. However, if a future refactor moves this line, it would crash. A defensive comment or variable capture would be clearer.

---

### Design / Architecture

**4. Accessing 6 private attributes via `noqa: SLF001` (Medium)**

`open_bandit_benchmark.py:239-268` — `build_policy_action_dist` reaches into `adapter._n_rounds`, `_n_actions`, `_item_feature_0`, `_item_popularity`, `_affinity`, `_position` — all private. This is the module's biggest coupling risk. If `BanditLogAdapter` refactors any of these internals, `build_policy_action_dist` silently breaks. Consider either:
- Adding a public method to `BanditLogAdapter` like `compute_action_dist(parameters)` that returns the `(n_rounds, n_actions)` array.
- Or adding public read-only properties for the arrays needed.

The noqa comments acknowledge this, but the right fix is to expose a proper API.

**5. `scripts/open_bandit_benchmark.py` test file uses `sys.path` manipulation (Low)**

`test_open_bandit_benchmark_script.py:15-17` — Inserting `scripts/` into `sys.path` to import the script is fragile. If another module named `open_bandit_benchmark` exists on the path, it would shadow. Consider making scripts a package or using `importlib`.

**6. `_sanitize_for_json` doesn't handle `set` or `frozenset` (Low)**

`scripts/open_bandit_benchmark.py:81-98` — If any downstream code accidentally passes a `set` or `frozenset`, the serializer won't convert it. Not currently triggered, but worth noting for robustness.

---

### Performance

**7. Reward model recomputed per null-control seed (Medium)**

`open_bandit_benchmark.py:494` — Every `run_strategy(..., null_control=True)` call computes `compute_reward_model(bf, seed=0)` where `bf` is the permuted feedback. Since the permutation seed is fixed across all null-control calls, the same `bf` (and therefore same reward model) is reconstructed for every (strategy, budget, seed) cell. That's 90 redundant reward-model fits on a 452K-row dataset. Consider caching the permuted reward model in the scenario or computing it once.

**8. Full `(n_rounds, n_actions)` array allocated per `build_policy_action_dist` call (Low)**

On 452K rows x 34 actions, this is ~120 MB per call. With 180 cells, that's a lot of temporary memory pressure. The array is short-lived, but if memory is ever a concern, the position-0-only path could allocate a smaller array.

---

### Test Coverage

**9. No test for `build_bandit_feedback_from_raw` error paths (Low)**

Tests cover the happy path and adapter validation, but there's no test for missing columns (`ValueError`), out-of-range action IDs, or the context-padding branch (`context.shape[1] < n_actions`). These are edge cases worth covering.

**10. No test for `_evaluate_gates` (Medium)**

The gate evaluation logic in the script is substantial (~160 lines) with multiple branches (null results present/absent, DR available/skipped, non-finite filtering). None of this is directly unit-tested. Consider extracting gate evaluation into a testable function (it already is `_evaluate_gates`, but no test calls it with synthetic data).

**11. `test_position_1_only_mask_restricts_rows` doesn't verify uniform mass values (Low)**

`test_open_bandit_benchmark.py:272-274` — The test checks that off-position rows sum to 1, but doesn't verify that each element equals `1/n_actions` (i.e., actually uniform). A softmax row also sums to 1.

**12. No test for `summarize_strategy_budget` filtering of null-control results (Low)**

The summary function filters `is_null_control` results, but no test verifies that null-control results are excluded from the summary.

---

### Code Quality

**13. Unused import in `__all__` (Low)**

`scripts/open_bandit_benchmark.py:65` imports `BanditLogAdapter` and re-exports it in `__all__` (line 556), but it's never used within the script itself. The comment says "re-exports so the CLI surface references land-of-truth imports" — but if nothing imports from the script, the re-export is dead.

**14. `_DEFAULT_POSITION_HANDLING_FLAG` constant vs adapter default (Low)**

`open_bandit_benchmark.py:70` defines `_DEFAULT_POSITION_HANDLING_FLAG = "position_1_only"`, matching the adapter's default. If the adapter's default ever changes, these would silently diverge. Consider importing the default from the adapter.

**15. Report helper hardcodes strategy order (Low)**

`_open_bandit_report_helper.py:186` — `strategies = ["random", "surrogate_only", "causal"]` is hardcoded. If a new strategy is added, the helper won't include it. Could derive from the data.

**16. Mutable default dict in `OpenBanditBenchmarkResult` (Low)**

`open_bandit_benchmark.py:373-375` — Using `field(default_factory=dict)` for `selected_parameters` and `diagnostics` is correct. The `selected_parameters: dict[str, Any] | None = None` default is also fine. No issue here; just noting it's done correctly.

---

### Numerical / Statistical

**17. Softmax tau floor differs between module and adapter (Low)**

`open_bandit_benchmark.py:255` uses `max(tau, 1e-6)` while the adapter uses `max(tau, _TAU_FLOOR)`. I'd want to verify `_TAU_FLOOR == 1e-6` — if they differ, the policies diverge at extreme temperatures.

**18. `compute_reward_model` doesn't account for position (Low-Medium)**

The reward model computes `E[reward | action=a]` ignoring position entirely. On Men/Random where position-specific CTR varies, this could bias the DM estimate. The code acknowledges this as intentional ("zero-context baseline" — line 297), and DR's cross-check passes within 1%, so it's not hurting verdict quality. But it's worth flagging for Sprint 36+ if richer reward models are planned.

---

### Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| Medium | 4 | Private attribute coupling (#4), gate DR alignment (#2), null-control reward model caching (#7), gate evaluation untested (#10) |
| Low | 10 | Various test coverage gaps, minor code quality items |

The PR is solid overall. The main structural concern is the tight coupling to `BanditLogAdapter` private attributes in `build_policy_action_dist`. The gate evaluation logic deserves its own unit tests given its complexity. Performance-wise, the redundant reward model computation during null-control runs is the most impactful optimization opportunity.

---

datablogin:
# Claude Code Review

## Review Feedback

Now I have a thorough understanding of the code. Here is my review:

---

## PR Review: Sprint 35.C Open Bandit Benchmark

Overall, this is a well-structured benchmark harness with strong provenance tracking, thorough test coverage of the primary module, and careful numerical handling. The code quality is high. Below are my findings organized by severity.

### Bugs / Correctness Issues

**1. `build_policy_action_dist` accesses 6 private attributes via `noqa: SLF001` — tight coupling risk**
`open_bandit_benchmark.py:239-268`

`build_policy_action_dist` reaches into `adapter._n_rounds`, `_n_actions`, `_item_feature_0`, `_item_popularity`, `_affinity`, `_position`. While the `noqa` comments acknowledge this, the function is essentially duplicating the adapter's internal softmax math. If `BanditLogAdapter.run_experiment` ever changes its scoring formula, this function will silently diverge. Consider adding a public method on the adapter (e.g., `adapter.compute_action_dist(parameters)`) that both `run_experiment` and `build_policy_action_dist` call, or at least a cross-reference test that asserts the two paths agree for several parameter combos. The existing `test_action_dist_matches_adapter_snipw` partially does this but only tests one parameter combo under `"marginalize"` mode — the `"position_1_only"` path has no equivalence test against the adapter.

**2. Null-control recomputes reward model on permuted data, but the real-data reward model is pre-cached**
`open_bandit_benchmark.py:493-494`

When `null_control=True`, `run_strategy` calls `compute_reward_model(bf, seed=0)` on the permuted feedback to get a new reward hat. This is correct for DM/DR on permuted data, but note that the reward model is deterministic (`del seed`), so calling it on the permuted `bf` produces the same per-action means as the original data (permutation preserves per-action marginals when stratified by position). The DR estimate on permuted data will thus use the same reward_hat as real data in expectation. This isn't a bug — the null control is about destroying the reward-to-row association, not the reward-to-action association — but worth calling out in a comment since it's subtle.

**3. `_sanitize_for_json` doesn't handle `set` or `frozenset` types**
`scripts/open_bandit_benchmark.py:81-98`

If any gate result or provenance dict ever contains a set (e.g., from a future gate), `json.dump` will raise `TypeError`. Low risk today since all dataclass fields are lists/dicts/scalars, but fragile going forward. Consider adding a `set`/`frozenset` → `sorted(list(...))` branch.

### Design & Maintainability

**4. Report helper uses `sys.path.insert` to import the script**
`test_open_bandit_benchmark_script.py:15-19`

The test file manipulates `sys.path` to import `scripts/open_bandit_benchmark.py` as a module. This is brittle — it breaks if pytest's working directory changes or if the scripts directory gets moved. A more robust approach would be to either:
- Make `scripts/` a package with `__init__.py` and import via the package path, or
- Use `importlib.util.spec_from_file_location` for explicit path-based import.

**5. `_evaluate_gates` uses `max(budgets)` as verdict budget, not a hard-coded `80`**
`scripts/open_bandit_benchmark.py:404`

The warning at line 411 notes this, but the function doesn't expose a `verdict_budget` parameter to let the caller override it. If someone passes `--budgets 20,40`, the gates silently run at B40 with a warning. A `--verdict-budget` CLI flag would make this explicit.

**6. Unused import in `scripts/open_bandit_benchmark.py`**
Line 65: `from causal_optimizer.domain_adapters.bandit_log import BanditLogAdapter` is imported and placed in `__all__` but never used in the script body. The `__all__` comment says it's a "re-export" but this is a script file, not a library module — re-exports from scripts are unusual and may confuse readers.

Similarly, `build_policy_action_dist` is imported and placed in `__all__` at line 60/569 but is not called anywhere in the script.

### Numerical / Statistical

**7. `compute_reward_model` clamps to `[0, 1]` — only valid for binary rewards**
`open_bandit_benchmark.py:320`

The `np.clip(action_mean, 0.0, 1.0)` assumes binary rewards. The docstring explains this, but if this module is reused for non-binary outcomes (e.g., revenue), the clamp would silently truncate. Consider making the clamp bounds parameters or removing the clamp if the numerical drift it guards against is truly negligible.

**8. `position_handling_flag` values other than `"position_1_only"` and `"marginalize"` are silently accepted**
`open_bandit_benchmark.py:263`

`build_policy_action_dist` only checks `if position_flag == "position_1_only"` and otherwise falls through. If someone passes a typo like `"position_one_only"`, the function silently uses the `marginalize` behavior. Consider adding validation or an explicit `elif position_flag == "marginalize"` with an `else: raise ValueError`.

### Test Coverage

**9. Missing tests for edge cases and negative paths**

Good coverage overall, but some gaps:
- **No test for `build_bandit_feedback_from_raw` with missing columns** — the function has validation logic (lines 110-121) but no test exercises the `ValueError` paths for missing data columns or missing `item_context` columns.
- **No test for `build_bandit_feedback_from_raw` with out-of-range action IDs** — the validation at line 146 is untested.
- **No test for `run_strategy` with `budget <= 0`** — the `ValueError` at line 484 is untested.
- **No test for `null_control=True` without `permutation_seed`** — the `ValueError` at line 486 is untested.
- **No test for `use_reward_model=False`** — verifying that DM and DR are `None` when the reward model is disabled.
- **`_open_bandit_report_helper.py` has zero test coverage** — the helper is a standalone script with substantial logic (MWU comparison, classification, formatting). Even a single smoke test that runs `main()` on a fixture artifact would catch regressions.

**10. Test for `position_1_only` doesn't verify that on-position rows have non-uniform distribution**
`test_open_bandit_benchmark.py:248-274`

The test checks that off-position rows sum to 1, but doesn't verify that on-position rows have the expected non-uniform softmax distribution. It also doesn't check that off-position rows are uniform (only that they sum to 1).

### Minor / Style

**11. `_DEFAULT_POSITION_HANDLING_FLAG` is defined but only used in one place**
`open_bandit_benchmark.py:70`

Consider inlining it as a default parameter value in `build_policy_action_dist` to reduce indirection.

**12. Duplicate `dict(best_metrics)` copy**
`open_bandit_benchmark.py:544`

`diagnostics: dict[str, float] = dict(best_metrics)` creates a shallow copy, but `best_metrics` is already a fresh dict created by the comprehension at line 518 or 535, so the `dict()` copy is redundant.

**13. `_load` in the report helper is trivially thin**
`_open_bandit_report_helper.py:22-24`

The function just wraps `json.loads(Path(...).read_text())` with a type annotation. It adds little value and could be inlined, though this is very minor.

### Report Quality

The benchmark report (`sprint-35-open-bandit-benchmark-report.md`) is thorough and well-structured with proper attribution, provenance, and scope boundaries. One note: the report includes a local filesystem path in the Artifacts section (`/Users/robertwelborn/Projects/_local/...`). If this report is intended for external consumption, consider redacting or parameterizing that path.

### Summary

This is solid, production-quality benchmark code. The main actionable items are:
1. Add a public `compute_action_dist` method to `BanditLogAdapter` to eliminate private-attribute coupling (item 1)
2. Validate `position_handling_flag` values explicitly (item 8)
3. Fill the test coverage gaps, especially for validation error paths and the report helper (item 9)
4. Clean up unused imports/re-exports in the script (item 6)

---

datablogin:
@greptileai

PR #198, opened 2026-04-22
sprint 37: Open Bandit prior graph + A1 minimal-focus rerun

datablogin:
## Summary

Lands the Sprint 36 preregistration as **Option A1** and reruns Men/Random.

- `BanditLogAdapter.get_prior_graph()` returns the preregistered seven-node / six-edge graph (every search variable directly parents `policy_value`, no bidirected edges).
- New `ExperimentEngine` flag `pomis_minimal_focus` (default `False`) opts a workload into the A1 minimal-focus heuristic: when on, the focus helper restricts to `screened ∩ ancestors` whenever (a) a graph is supplied, (b) ancestors cover the full search space, and (c) the resulting intersection is a non-empty proper subset of the search space. Applied in both optimization and exploitation so B80 (which crosses the `>= 50` exploitation boundary) does not silently revert.
- The Open Bandit benchmark harness sets the flag only on the `causal` arm; `surrogate_only` and `random` are mechanically identical to Sprint 35.

## Sprint 37 Verdict (Men/Random, full slice, 10 seeds)

- **B80 `causal` vs `surrogate_only`:** `p = 0.7337` two-sided MWU — **near-parity** (means agree to 6 decimals: 0.006181 vs 0.006182). H0 from the Sprint 36 plan is confirmed.
- B40: `p = 0.2123` not significant. B20: `p = 0.0820` trending (mean direction `causal < surrogate_only`; not certified).
- Both optimized strategies still **certified** over `random` at every budget (`p = 0.0002`).
- Sprint 35's bit-identical `causal == surrogate_only` tie is broken on every seed at every budget — A1 changes the trajectory exactly as Sprint 36's path-by-path engine analysis (path 4: soft-causal reranker) predicted, but the verdict-budget mean does not move.
- All five Section 7 gates **PASS**. Backend is `ax_botorch` on every cell with no RF fallback.

Full report: [`thoughts/shared/docs/sprint-37-open-bandit-prior-graph-report.md`](thoughts/shared/docs/sprint-37-open-bandit-prior-graph-report.md).

Closes #197.

## Test plan

- [ ] Verify the preregistered graph in `BanditLogAdapter.get_prior_graph()` matches the Sprint 36 plan node-by-node and edge-by-edge.
- [ ] Confirm `pomis_minimal_focus` defaults to `False` so non-Open-Bandit workloads are unchanged.
- [ ] Confirm the benchmark harness wires the flag only on `causal` (`surrogate_only` and `random` unchanged).
[...]

PR #202, opened 2026-04-23
docs: sprint 38 post-A1 recommendation (option B)

### LINKED ISSUE #201: Sprint 38: recommend the post-A1 Open Bandit follow-up (Option B, C, or D)
## Context

Sprint 37 closed the Sprint 36 preregistration loop by landing Option A1 and rerunning Open Bandit Men/Random on the full slice.

Merged result (PR #198 / issue #197):

- the first preregistered Open Bandit prior graph is now implemented
- the `pomis_minimal_focus` flag is wired only on the `causal` arm
- Sprint 35's exact bit-identical tie is broken on every seed
- B80 `causal` vs `surrogate_only` remains near-parity (`p = 0.7337`; means `0.006181` vs `0.006182`)
- both optimized strategies remain certified over `random`
- all five Section 7 gates pass
- backend provenance is Ax/BoTorch only, with no RF fallback mixing

Sprint 37 therefore answered the A1 question cleanly: the heuristic changes the trajectory mechanically, but it does **not** produce a certified causal advantage on Men/Random. The Sprint 37 report says Sprint 38 should now pick exactly one follow-up:

1. **Option B (graph widening):** add one non-ancestor structural node so `_get_focus_variables` returns a proper subset directly
2. **Option C (different heuristic):** keep the graph but replace A1 with a materially different, preregisterable restriction rule
3. **Option D (move on):** accept that under the Sprint 35 surface the two paths converge and reopen another lane instead

## Scope (planning only — no implementation)

1. Write `thoughts/shared/plans/27-sprint-38-recommendation.md`.
2. Write `thoughts/shared/prompts/sprint-38-open-bandit-post-a1-follow-up.md` as the execution prompt for the next implementation lane.
3. Answer the central Sprint 38 question: after A1 near-parity, what is the single highest-value next move?

## Concrete recommendation must include

1. Choose **exactly one** of Option B, C, or D.
2. Explain why the chosen lane is more informative than the two rejected lanes.
3. State whether the next move fits in one implementation issue or needs two ordered issues.

If the recommendation chooses **Option B**, it must also include:

- the exact added structural node or nodes
- the exact directed edges to add or re-route
- code-grounded justification from `bandit_log.py` for every new node/edge
- confirmation that no bidirected edges are introduced
- the exact falsifiable rerun outcome to test next

If the recommendation chooses **Option C**, it must also include:

- the exact heuristic to test
- the exact engine surfaces it would touch
- why it is not a post-hoc chase of the non-certified B20 trend
- the exact falsifiable rerun outcome to test next

If the recommendation chooses **Option D**, it must also include:

- the exact criterion for declaring the Open Bandit A1 lane exhausted for now
- which lane reopens next instead (for example second dataset, multi-objective, or another already-queued frontier)
- why moving on is more honest than another Open Bandit iteration right now

## Empirical setup to keep locked if Sprint 38 stays on Open Bandit

- same Men/Random full slice
- same budgets `20, 40, 80`
- same seeds `0..9`
- same SNIPW-primary verdict rule
- same Section 7 gates
- same null-control permutation seed `20260419`
- same Ax/BoTorch-only provenance rule for verdict rows

## What not to propose next

- do not chase the B20 trend as if it were certified
- do not open a power-extension rerun for A1; Sprint 37 already showed the B80 outcome is unambiguous
- no Women / All / BTS-primary
- no slate OPE
- no DRos-primary
- no second dataset unless the recommendation explicitly chooses Option D and justifies the switch
- no bidirected-edge widening unless the recommendation makes a fresh, code-grounded case for reopening that deferred scope
- do not combine multiple follow-ups into one sprint

## Standards

- The recommendation must be honest about the Sprint 37 result: A1 changed the path, not the verdict.
- Option B should be treated as the default baseline to beat, because it is the next structurally distinct Open Bandit test after A1.
- Option C bears a higher burden of proof than Option B; another heuristic tweak must show why it is not just post-hoc tuning.
- Option D is acceptable if and only if it is argued directly from the evidence rather than from impatience or queue pressure.
- Keep the next implementation lane small enough to fit in a single focused PR if possible.

## Workflow

`/tdd -> implement -> /polish -> gh pr create -> /gauntlet`. Do not merge; leave the PR open for human review.

datablogin:
## Summary

- Picks **Option B** for Sprint 38: widen the Open Bandit preregistered prior graph with one non-search-space structural node (`logged_position_distribution`) and one directed edge (`logged_position_distribution -> position_handling_flag`).
- Preserves the Sprint 37 engine surface verbatim: no edits to `suggest.py`, `loop.py`, `open_bandit.py`, or `open_bandit_benchmark.py`. `pomis_minimal_focus` stays default-off at the engine level and enabled only on the `causal` arm.
- Preregisters H0 (B80 `p > 0.15`, recapitulating Sprint 37 near-parity) as the *predicted* outcome and makes that no-op prediction the falsifiable claim — a B80 certified or trending row in either direction would prove the Sprint 37 engine-path analysis incomplete.
- Rejects **Option C** (different heuristic) because it would tune against the non-certified Sprint 37 B20 trend (`p = 0.0820`, `causal < surrogate_only` on 8/10 seeds). Rejects **Option D** (move on) because the Open Bandit A1 lane has not been stress-tested along the graph-structure axis after exactly one shape.
- Recommends one Sprint 38 implementation issue (not two): a single-edge graph widening plus a Sprint 37-shape rerun, strictly smaller than the Sprint 37 PR #198 footprint.

## Terminology note

The Sprint 36 plan and Sprint 38 orchestration prompt used the shorthand "non-ancestor structural node." The recommendation retires that phrase: once the edge `logged_position_distribution -> position_handling_flag` is added, `logged_position_distribution` *is* an ancestor of `policy_value` via the two-hop path. The accurate term is "non-search-space structural node" — the A1 binding condition (`len(ancestors_in_space) == len(all_var_names)`) holds because the new node is excluded from `_ancestors_in_space` by the `search_space.variable_names` filter, not because it is a non-ancestor. See the recommendation's Sprint Theme section for the full note.

## Files added

- `thoughts/shared/plans/27-sprint-38-recommendation.md` — the recommendation, grounded in `causal_optimizer/domain_adapters/bandit_log.py` lines 364–367 within `run_experiment` 346–369 and `get_prior_graph` 513–541 for the node/edge justification, and in Sprint 37 B80 `p = 0.7337` / B20 `p = 0.0820` for the option ranking.
- `thoughts/shared/prompts/sprint-38-open-bandit-post-a1-follow-up.md` — execution prompt for the next implementation lane, following the orchestration spec's `/tdd → implement → /polish → gh pr create → /gauntlet → report` workflow.

## TDD note (docs-only adaptation)

Per the Sprint 38 orchestration prompt for Track B, TDD for a planning-doc track is satisfied by defining an acceptance checklist the recommendation must satisfy before writing it. The checklist:

- picks exactly one option (B, C, or D)
- references Sprint 37 numbers, not queue pressure
- if B: names node, edge, code-grounded line citations, falsifiable rerun outcome
- does not chase the B20 trend
- does not propose a power-extension rerun for A1
- names one-issue-or-two sprint shape
- ships a matching execution prompt that does not silently widen scope

The two artifacts were written against that spec; both satisfy every item. No Python tests written.

## Gauntlet history

- `/greploop` — iteration 1 cleared 2 P2 comments (Sprint Theme self-contradiction + "non-ancestor" misnomer); re-review posted zero new inline comments.
- `/claudeloop` — 5 iterations (reached the max). Each iteration address real actionable items: `suggest.py` line-number drift (iter 1), prompt citation drift (iter 2), path-1/path-4 precision + scope dual-condition + concrete test edits (iter 3), test-file positive assertion for the new edge (iter 4), null-result asymmetry + drift-guard banner (iter 5). Iteration 5 explicitly said "minor nits above are polish, not blockers. Recommend merge after human sign-off."
- `/check-pr` — all 9 CI checks pass (lint, typecheck, build, test 3.10–3.13, test-bayesian, codecov/patch). Zero unresolved review threads.

## Test plan

- [ ] Reviewer confirms Option B is the named pick and the Option C / Option D rejections cite Sprint 37 evidence rather than impatience.
- [ ] Reviewer confirms the added node `logged_position_distribution` and the added edge `logged_position_distribution -> position_handling_flag` are each grounded in a specific `bandit_log.py` line range (364–367 within `run_experiment` 346–369 is cited in both the plan and the prompt).
- [ ] Reviewer confirms no bidirected edges are proposed.
- [ ] Reviewer confirms the falsifiable H0 / H1 / H2 block is preregistered and not tuned against an unpublished observation.
- [ ] Reviewer confirms the engine-path no-op prediction's dual scope conditions (new node non-search-space; new edge does not add a parent to `policy_value`) are stated explicitly so a Sprint 39+ plan cannot transplant the prediction to a different widening.
- [ ] Reviewer confirms the prompt's "do not" list forbids engine/optimizer edits, `request_item_overlap` (the deferred second candidate), second-dataset reopens, power extensions on A1, and B20 chasing.
- [ ] Reviewer confirms the Sprint 39 next-move branch map is preregistered (four outcomes → four Sprint 39 shapes).

Closes #201.

Planning PR only — no code changes. Implementation lands in a separate Sprint 38 PR after this planning PR is human-approved and merged.

### DIFF
--- a/thoughts/shared/plans/27-sprint-38-recommendation.md
+++ b/thoughts/shared/plans/27-sprint-38-recommendation.md
@@ -4,10 +4,35 @@ Updated: 2026-04-22
 
 ## Sprint Theme
 
-**Option B — graph widening. Add one non-ancestor structural node
-(`logged_position_distribution`) to the Open Bandit prior graph so
-`_get_focus_variables` returns a real proper subset directly, instead
-of relying on the A1 screening intersection to do the restriction.**
+**Option B — graph widening. Add one non-search-space structural node
+(`logged_position_distribution`) to the Open Bandit prior graph. The
+widening is a preregistered structural test of the Sprint 37
+engine-surface analysis — confirmed below as a predicted no-op along
+every engine path that already existed under the Sprint 37
+`pomis_minimal_focus=True` wiring. Under the widened graph,
+`_get_focus_variables` still returns the full search space (the new
+node is not a search-space variable), `_apply_minimal_focus_a1` still
+binds and still returns the same `screened ∩ ancestors` intersection
+as Sprint 37, and the soft-causal reranker's alignment score is
+unchanged. The no-op prediction *is* the falsifiable claim — a B80
+certified or trending row in either direction would prove the
+Sprint 37 engine-path analysis incomplete.**
+
+**Terminology note.** The Sprint 38 orchestration prompt and the
+Sprint 36 plan both used the phrase "non-ancestor structural node"
+as shorthand for "a structural node that is not itself a
+search-space knob." That shorthand is graph-theoretically imprecise
+once the edge
+`logged_position_distribution -> position_handling_flag` is added,
+because `logged_position_distribution` then *is* an ancestor of
+`policy_value` via the two-hop path through `position_handling_flag`.
+To keep the engine-path reasoning unambiguous, this recommendation
+uses **"non-search-space structural node"** (or simply "structural
+node") instead of "non-ancestor" from this point forward. The A1
+binding condition holds under the widened graph precisely because
+`logged_position_distribution` is not a search-space variable
+(`_ancestors_in_space` filters it out of its search-space
+intersection), not because it is a non-ancestor of the objective.
 
 Sprint 37 landed Option A1 cleanly (PR #198). The preregistered
 seven-node / six-edge graph is now live in
@@ -25,16 +50,17 @@ explicitly say not to chase. See
 
 Sprint 38 picks exactly one of the three follow-ups the Sprint 37
 report named. The pick is **Option B** — graph widening with one
-non-ancestor structural node, because it is the next structurally
-distinct test of whether the Men/Random slice can separate `causal`
-from `surrogate_only` under the verdict rule. Options C and D are
-rejected below, on the evidence, not queue pressure.
+non-search-space structural node, because it is the next
+structurally distinct test of whether the Men/Random slice can
+separate `causal` from `surrogate_only` under the verdict rule.
+Options C and D are rejected below, on the evidence, not queue
+pressure.
 
 ## Goal
 
 Sprint 38 should end with:
 
-1. one new non-ancestor structural node
+1. one new non-search-space structural node
    (`logged_position_distribution`) added to the preregistered graph
    returned by `BanditLogAdapter.get_prior_graph()`, with a single
    directed edge `logged_position_distribution -> position_handling_flag`
@@ -49,8 +75,10 @@ Sprint 38 should end with:
    `0.05 < p <= 0.15` at B80 — same two-sided MWU convention as
    Sprint 35 and Sprint 37
 4. a Sprint 38 report that answers one question end-to-end: does
-   graph widening change the B80 verdict or does it recapitulate the
-   Sprint 37 near-parity under a genuinely restricted focus set?
+   pure graph widening (one new non-search-space ancestor of
+   `policy_value`, same search space, same engine surface) move the
+   B80 verdict, or does it recapitulate the Sprint 37 near-parity as
+   the engine-path analysis predicts?
 
 No second node in Sprint 38. No bidirected edges. No new slice. No
 DRos-primary. No second dataset. No new heuristic. No power extension
@@ -58,7 +86,7 @@ on A1. The scope is one node, one edge, one rerun, one verdict row.
 
 ## Why This Sprint Now
 
-### Why Option B (this sprint, Men/Random, one non-ancestor node)
+### Why Option B (this sprint, Men/Random, one non-search-space structural node)
 
 Under the preregistered graph Sprint 37 landed, every search variable
 is already an ancestor of `policy_value`. That means
@@ -74,13 +102,21 @@ produce a different trajectory (the Sprint 35 exact tie is broken)
 but does not move the verdict-budget mean.
 
 The natural next test is to add structure to the graph itself —
-specifically, one non-ancestor node that makes the ancestor DAG
-non-trivial. This is Option B in the Sprint 36 plan and in the
-Sprint 37 "Sprint 38+ Implications" section. The Sprint 36 plan's
-"Exit Criterion" section names two candidate structural nodes:
-`logged_position_distribution` and `request_item_overlap`. Sprint 38
-commits to exactly one: `logged_position_distribution` with a
-directed edge `logged_position_distribution -> position_handling_flag`.
+specifically, one non-search-space structural node that makes the
+ancestor DAG non-trivial (some search-space variables gain
+non-search-space parents; the objective gains an ancestor that is
+not itself a knob). This is Option B in the Sprint 36 plan and in
+the Sprint 37 "Sprint 38+ Implications" section. The Sprint 36
+plan's "Exit Criterion" section names two candidate structural
+nodes: `logged_position_distribution` and `request_item_overlap`.
+Sprint 38 commits to exactly one: `logged_position_distribution`
+with a directed edge
+`logged_position_distribution -> position_handling_flag`. (The
+Sprint 36 plan and the Sprint 38 orchestration prompt called these
+"non-ancestor" nodes; after adding the two-hop edge,
+`logged_position_distribution` is in fact an ancestor of
+`policy_value`. The accurate term is "non-search-space" — see the
+Sprint Theme terminology note above.)
 
 **Honest prediction about what this widening changes.** Under the
 Sprint 37 engine surface, the widened graph is predicted to be a
--- a/thoughts/shared/plans/27-sprint-38-recommendation.md
+++ b/thoughts/shared/plans/27-sprint-38-recommendation.md
@@ -246,10 +246,10 @@ Sprint 38 adds exactly one new node and exactly one new directed edge.
 8. `logged_position_distribution`
 
 **Code-grounded justification.** The adapter's `run_experiment`
-branches on the `position_handling_flag` categorical to pick the
-logged-row subset that enters the SNIPW sum
 ([`causal_optimizer/domain_adapters/bandit_log.py`](../../causal_optimizer/domain_adapters/bandit_log.py)
-lines 352–369):
+lines 346–369) branches on the `position_handling_flag` categorical
+to pick the logged-row subset that enters the SNIPW sum. The
+row-mask branch itself is at lines 364–367:
 
 ```python
 if position_flag == "position_1_only":
@@ -278,12 +278,13 @@ bullet) and the Sprint 37 report's "Sprint 38+ Implications" section.
 
 **Code-grounded justification.** The flag's behavior is a direct
 function of `self._position`
-(`bandit_log.py` lines 352–369, cited above). Per the Sprint 36
-plan's edge-justification discipline, each added edge must cite a
-specific line that makes the upstream variable structural. The line
-range above — the row-mask branch inside `run_experiment` — is the
-site that makes `logged_position_distribution` structurally upstream
-of `position_handling_flag`.
+(`bandit_log.py` lines 364–367, cited above, inside
+`run_experiment` at lines 346–369). Per the Sprint 36 plan's
+edge-justification discipline, each added edge must cite a
+specific line that makes the upstream variable structural. The
+row-mask branch is the site that makes
+`logged_position_distribution` structurally upstream of
+`position_handling_flag`.
 
 No edge from `logged_position_distribution` directly to `policy_value`
 is added. The flow is:
@@ -685,8 +686,9 @@ that forced the choice.
 - Sprint 36 preregistration plan: [26-sprint-36-recommendation.md](26-sprint-36-recommendation.md)
 - Sprint 37 Open Bandit prior-graph report: [sprint-37-open-bandit-prior-graph-report.md](../docs/sprint-37-open-bandit-prior-graph-report.md)
 - `BanditLogAdapter`: `causal_optimizer/domain_adapters/bandit_log.py`
-  (see `get_prior_graph`, lines 513–541, and `run_experiment` row-mask
-  branch, lines 352–369, for the Option B node justification)
+  (see `get_prior_graph`, lines 513–541, and `run_experiment`
+  row-mask branch, lines 364–367 within `run_experiment` 346–369,
+  for the Option B node justification)
 - Open Bandit OPE stack and Section 7 gates: `causal_optimizer/benchmarks/open_bandit.py`
 - Open Bandit benchmark runner: `causal_optimizer/benchmarks/open_bandit_benchmark.py`
   (enables `pomis_minimal_focus` only on the `causal` arm, line 541)
--- a/thoughts/shared/plans/27-sprint-38-recommendation.md
+++ b/thoughts/shared/plans/27-sprint-38-recommendation.md
@@ -510,6 +510,21 @@ seeds. A Sprint 39 power-extension rerun is the explicit answer to
 that risk if Sprint 38 produces a trending row (H2); **Sprint 38
 itself does not extend power on A1 or on Option B**.
 
+**Null-result asymmetry.** H0 is also 10-seed-gated. If the true
+effect is null *and* the 10-seed test would be underpowered to
+certify a small true effect, H0 passes — but it can pass for two
+very different reasons: (a) the widening is genuinely a no-op along
+every engine path (the prediction), or (b) the widening produces a
+real but sub-power effect that 10 seeds cannot distinguish from a
+no-op. Sprint 38 partially guards against (b) by treating B20/B40
+near-parity *alongside* B80 near-parity as the exhaustion
+criterion. The Sprint 38 report's reconciliation section must
+address this asymmetry explicitly: a B80 near-parity that is paired
+with B20 and B40 near-parity reads as genuine no-op (path to
+Sprint 39 Option D); a B80 near-parity paired with non-trivial B20
+or B40 movement reads as a potential sub-power miss and should be
+flagged for Sprint 39 diagnostic triage rather than Option D.
+
 ## Recommended Issue Shape
 
 **One issue for Sprint 38.** The scope is a single graph widening
--- a/thoughts/shared/plans/27-sprint-38-recommendation.md
+++ b/thoughts/shared/plans/27-sprint-38-recommendation.md
@@ -340,23 +340,34 @@ engine or optimizer change):
    ancestor set inside the search space is unchanged, so the A1
    result is the same set the Sprint 37 rerun computed at every
    (seed, budget, phase) triple.
-3. Parent-weighted exploitation (`suggest.py` lines 569–586): uses
-   `causal_graph.parents(objective_name)`. Sprint 37 graph gave
-   `parents(policy_value) = 6` search-space vars → `weights = None`
-   (no restriction). Sprint 38 widened graph still gives
-   `parents(policy_value) = 6` search-space vars (the new edge
-   targets `position_handling_flag`, not `policy_value`) → still
-   `weights = None`, still no-op.
-4. Soft-causal reranker (`_rerank_alignment_only` +
-   `_causal_alignment_score`, `suggest.py` lines 751–812 and
-   982–1021): averages over ancestors of the objective. The
-   ancestor set grows by one (`logged_position_distribution`), but
-   that variable is not in the search space, so
-   `_causal_alignment_score` skips it when it iterates over
-   ancestors in the search space (the per-variable normalized
-   displacement loop at lines 1010–1019 only fires when the
-   ancestor name is a key in `candidate` and `best_params`). The
-   reranker's per-seed output is therefore identical to Sprint 37.
+3. Parent-weighted exploitation (`suggest.py` lines 604–619, inside
+   `_suggest_exploitation` which starts at line 532): uses
+   `causal_graph.parents(objective_name)`. The 70/30 weighting
+   activates only when `parent_focus` is a *proper* subset of
+   `eligible_vars` (line 608:
+   `if parent_focus and len(parent_focus) < len(eligible_vars)`).
+   Sprint 37 graph gave `parents(policy_value) = 6` search-space
+   vars → `weights = None` (no restriction). Sprint 38 widened
+   graph still gives `parents(policy_value) = 6` search-space vars
+   (the new edge targets `position_handling_flag`, not
+   `policy_value`) → still `weights = None`, still no-op.
+4. Soft-causal reranker (`_rerank_alignment_only` at `suggest.py`
+   lines 848–877, called from `_suggest_bayesian` at line 839; soft
+   path entered via `use_soft = causal_graph is not None and
+   causal_softness < _HARD_FOCUS_THRESHOLD` at line 784, with
+   ancestor set built at lines 810–813; alignment score function
+   `_causal_alignment_score` at lines 1015–1054): averages over
+   ancestors of the objective. The ancestor set grows by one
+   (`logged_position_distribution`), but that variable is not in
+   the search space, so `_causal_alignment_score` skips it in the
+   per-variable normalized-displacement loop at lines 1043–1052 —
+   the filter is search-space membership
+   (`var = var_map.get(name); if var is None: continue` at lines
+   1044–1046, where `var_map = {v.name: v for v in search_space.variables}`
+   at line 1040), not key-presence in `candidate`/`best_params`
+   (those fall back to `mid` via `candidate.get(name, mid)` at line
+   1050). The reranker's per-seed output is therefore identical to
+   Sprint 37.
 5. `_suggest_causal_gp`: inert (not requested by the benchmark).
 6. `causal_exploration_weight`: pinned to `0.0`, inert.
 
@@ -685,10 +696,16 @@ that forced the choice.
     (`_get_focus_variables`)
   - `causal_optimizer/optimizer/suggest.py` lines 1183–1236
     (`_apply_minimal_focus_a1`, Sprint 37 A1 helper)
-  - `causal_optimizer/optimizer/suggest.py` lines 751–812
-    (soft-causal reranker inside `_suggest_bayesian`)
-  - `causal_optimizer/optimizer/suggest.py` lines 982–1021
-    (`_causal_alignment_score`)
+  - `causal_optimizer/optimizer/suggest.py` lines 604–619
+    (parent-weighted 70/30 split inside `_suggest_exploitation`,
+    which starts at line 532)
+  - `causal_optimizer/optimizer/suggest.py` lines 848–877
+    (`_rerank_alignment_only`, called from `_suggest_bayesian` at
+    line 839 when `use_soft` is true at line 784)
+  - `causal_optimizer/optimizer/suggest.py` lines 1015–1054
+    (`_causal_alignment_score`; per-variable normalized-displacement
+    loop at 1043–1052 with search-space-membership filter at
+    1044–1046)
 - `CausalGraph` type: `causal_optimizer/types.py`
 - `DomainAdapter.get_prior_graph` contract:
   `causal_optimizer/domain_adapters/base.py`
--- a/thoughts/shared/plans/27-sprint-38-recommendation.md
+++ b/thoughts/shared/plans/27-sprint-38-recommendation.md
@@ -331,8 +331,11 @@ engine or optimizer change):
    `{v in search_space.variable_names | v in ancestors(policy_value)}`
    still equals the full search space, because
    `logged_position_distribution` is **not** itself a search-space
-   variable. So path 1 on its own remains a no-op — the full
-   search-space fallback triggers.
+   variable. `_get_focus_variables` returns that full-search-space
+   intersection verbatim (the intersection is truthy, so the
+   `focus if focus else search_space.variable_names` guard at line
+   1180 does not trigger a fallback — it just returns the already-equal
+   set). Path 1 on its own is therefore a no-op.
 2. A1 minimal-focus (`_apply_minimal_focus_a1`, `suggest.py` lines
    1183–1236): the binding condition — every search-space variable
    is an ancestor of `policy_value` — still holds, so the A1 helper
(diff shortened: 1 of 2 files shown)


datablogin:
@greptileai

datablogin:
# Claude Code Review

## Review Feedback

## PR Review: docs-only Sprint 38 recommendation

This is a docs-only PR (2 files, 983 lines) laying out a preregistered plan for Sprint 38 — one new non-search-space structural node (`logged_position_distribution`) and one directed edge added to the Open Bandit prior graph, then a Sprint 37-shape rerun. Overall the plan is clear, the no-op prediction is honest and explicitly falsifiable, and the terminology note at the top resolves the "non-ancestor" vs "non-search-space" ambiguity cleanly. A few issues worth addressing before implementation spins up.

### Line-number drift in `suggest.py` citations (plan, "What the engine would do with the widened graph, path by path")

The plan's per-path engine analysis is sound, but several cited line ranges don't match the current file. I verified against `causal_optimizer/optimizer/suggest.py` in this worktree.

| Doc claim | Cited range | Actual location |
|---|---|---|
| Parent-weighted exploitation | `suggest.py:569-586` | Lines 569-586 are the **categorical sweep** (`_suggest_exploitation`). The parent-weight / 70-30 split block is at **lines 604-619**. |
| Soft-causal reranker inside `_suggest_bayesian` | `suggest.py:751-812` | `_rerank_alignment_only` is defined at **lines 848-877**. The cited 751-812 range is the `_suggest_bayesian` docstring + soft-mode setup; the re-ranker itself is called at line 839 and lives further down. |
| `_causal_alignment_score` | `suggest.py:982-1021` | Function definition starts at **line 1015** (extends to 1054). Lines 982-1014 are the soft-reranking inside `_suggest_surrogate` (the RF surrogate path), not `_causal_alignment_score`. |
| "per-variable normalized displacement loop" | `suggest.py:1010-1019` | The loop is at **lines 1043-1052** inside `_causal_alignment_score`. |

Accurate citations in both files: `bandit_log.py:181`, `:195`, `:352-369`, `:513-541`, `suggest.py:1170-1180` (`_get_focus_variables`), `suggest.py:1183-1236` (`_apply_minimal_focus_a1`), `open_bandit_benchmark.py:541`. Only the four ranges above are off, but they occur in exactly the section the implementation PR must not alter — so future readers cross-referencing the claim will hit the wrong code.

### Subtle mechanism inaccuracy in path 4 (soft-causal reranker)

Plan (`27-sprint-38-recommendation.md`, path 4) says:

> `_causal_alignment_score` skips it when it iterates over ancestors in the search space (the per-variable normalized displacement loop at lines 1010–1019 only fires when the ancestor name is a key in `candidate` and `best_params`).

That's not quite the filter. The actual code (`suggest.py:1043-1052`) keys off search-space membership:

```python
for name in sorted(ancestor_names):
    var = var_map.get(name)                    # var_map = {v.name: v for v in search_space.variables}
    if var is None or var.lower is None or var.upper is None:
        continue
    ...
    c_norm = _normalize_value(var, float(candidate.get(name, mid)))  # defaults to mid
    b_norm = _normalize_value(var, float(best_params.get(name, mid)))
```

The filter is "`name` is in the search space" (`var_map.get(name) is not None`), not "`name` is a key in `candidate`/`best_params`" — the `.get(name, mid)` calls fall back to `mid` rather than skipping. The plan's conclusion holds (`logged_position_distribution` is not in search-space, so it's filtered), but the mechanism description is wrong and should be corrected to avoid misleading a future reader debugging a related path.

### Coupling risk for the widened `get_prior_graph`

Current `get_prior_graph` (`bandit_log.py:540`) programmatically enumerates the search space:

```python
edges = [(name, "policy_value") for name in self.get_search_space().variable_names]
```

The Sprint 38 widened graph cannot use this pattern — it needs an explicit edge list that includes the `logged_position_distribution -> position_handling_flag` edge, whose source is not a search-space variable. Worth flagging in the prompt's implementation section that the `get_prior_graph` body must switch to an explicit edge list (not a list comprehension over `variable_names`), and that unit tests should pin edges by name tuple rather than counting search-space membership. The existing test file `tests/unit/test_bandit_log_prior_graph.py` — which the prompt's item (2) updates — will need to pin this distinction explicitly.

### Plan/prompt duplication

The "Hard constraints" block in the prompt (15 items) and "What Sprint 38 Must Not Do" in the plan (15 items) are near-verbatim copies. The prompt acknowledges this with "any conflict … is resolved in favor of the recommendation," which is the right call, but every edit to the plan now requires a mirrored edit to the prompt. Consider either (a) collapsing the prompt's list to a pointer + delta, or (b) adding a commit-time check that the two stay aligned. Not a blocker.

### Verified factual claims

- Sprint 37 quoted results (`p = 0.7337` at B80, means 0.006181 vs 0.006182, B20 `p = 0.0820` with 8/10 seeds) all match `thoughts/shared/docs/sprint-37-open-bandit-prior-graph-report.md`.
- Graph shape claim (7 nodes / 6 edges for Sprint 37; 8 nodes / 7 edges under Option B) matches the six-variable search space in `bandit_log.py:313-344`.
- `pomis_minimal_focus=(strategy == "causal")` wiring at `open_bandit_benchmark.py:541` verified.
- Sprint 37 report (line 287) does name `logged_position_distribution` and `request_item_overlap` as Option B candidates, consistent with the plan.
- Permutation seed `20260419` is the same seed used in the Sprint 35 and Sprint 37 reports.
- Parents-of-`policy_value` argument in path 3 is correct: the new edge targets `position_handling_flag`, so `causal_graph.parents("policy_value")` is unchanged, leaving `weights = None` in the 70/30 split branch.

### What's good

- H0/H1/H2 are clearly falsifiable and the prediction is stated *before* the experiment runs.
- The terminology note resolves the "non-ancestor" vs "non-search-space" issue upfront — a real improvement over the Sprint 36 plan wording.
- Rejection arguments for Options C and D are evidence-based (B20 trend, structural priority, power discussion) rather than queue-pressure framing.
- Hard constraints list is comprehensive and mirrors the Sprint 34 contract discipline on bidirected edges and engine-surface changes.
- The "What Happens After Sprint 38" branch coverage (4 cases) is complete and each branch names a concrete next sprint move.

### Summary

Non-blocking issues — this is planning docs, not code. Recommend fixing the four `suggest.py` line ranges and the one mechanism-description sentence before the implementation agent reads the plan, since the engine-path no-op argument is the core of the sprint's value and a future reader will follow those line numbers. Also consider adding one sentence to the prompt's item (1) warning that the `get_prior_graph` implementation must switch from the current list comprehension to an explicit edge list.

---

datablogin:
# Claude Code Review

## Review Feedback

## Review: Sprint 38 Recommendation + Orchestration Prompt

This is a docs-only PR (no Python code touched), so most of the requested review categories (numpy/scipy APIs, dataclass usage, etc.) don't apply. I focused on what does matter for a planning doc whose central claim — "the widened graph is a no-op along every existing engine path" — depends entirely on accurate code citations.

### Summary

**Recommend: approve with one fix.** The recommendation's per-path engine analysis is rigorous and code-grounded, and the H0/H1/H2 framing is genuinely falsifiable. I verified every line citation in the recommendation against the actual code; all match. The orchestration prompt has one citation drift that should be corrected before an implementation agent starts work.

### Issues

**1. Line-number drift in the orchestration prompt** (must fix)

`thoughts/shared/prompts/sprint-38-open-bandit-post-a1-follow-up.md:147` cites:
> `_causal_alignment_score` (lines 982–1021)

But `causal_optimizer/optimizer/suggest.py:1015` is the actual `def _causal_alignment_score`, running to line 1054. The recommendation correctly cites `1015–1054`. Since the recommendation is declared authoritative ("any conflict between the two documents is resolved in favor of the recommendation"), an implementation agent following the prompt would land at the wrong code site. Easy fix: change `(lines 982–1021)` to `(lines 1015–1054)`.

**2. Minor citation imprecision** (optional)

`27-sprint-38-recommendation.md` cites `bandit_log.py` lines 352-369 as "the row-mask branch." The actual branch is `bandit_log.py:364-367`. Lines 352-363 are parameter parsing and validation. Not wrong, but a tighter cite (or `bandit_log.py:364-367 within run_experiment 346-369`) makes the structural-edge justification crisper.

### What I verified (all correct)

- `bandit_log.py:181` — `propensity_schema` ClassVar ✓
- `bandit_log.py:195` — `self._position` ✓
- `bandit_log.py:513-541` — `get_prior_graph` ✓
- `bandit_log.py:540` — list-comprehension pattern that the prompt correctly flags as needing replacement ✓
- `suggest.py:1170-1180` — `_get_focus_variables` ✓
- `suggest.py:1183-1236` — `_apply_minimal_focus_a1` ✓
- `suggest.py:1155-1167` — `_ancestors_in_space` filters by `search_space.variable_names`, confirming the no-op claim mechanically ✓
- `suggest.py:532` — `_suggest_exploitation` start ✓
- `suggest.py:604-619` — parent-weighted block, with the `len(parent_focus) < len(eligible_vars)` proper-subset gate at 608 ✓
- `suggest.py:728` — `_suggest_bayesian` start ✓
- `suggest.py:784` — `use_soft` check; `_HARD_FOCUS_THRESHOLD = 1e5` at line 77, so `causal_softness=0.5` does enter the soft path ✓
- `suggest.py:810-813` — ancestor-set construction ✓
- `suggest.py:839` — `_rerank_alignment_only` call ✓
- `suggest.py:848` — `_rerank_alignment_only` definition ✓
- `suggest.py:1015-1054` — `_causal_alignment_score` ✓
- `suggest.py:1040`, `1043-1052`, `1044-1046`, `1050` — sub-citations for `var_map`, the loop, the search-space-membership filter, and `candidate.get(name, mid)` ✓
- `open_bandit_benchmark.py:541` — `pomis_minimal_focus=(strategy == "causal")` ✓
- `tests/unit/test_bandit_log_prior_graph.py:71` — `assert len(graph.edges) == 6`; the prompt's guidance to switch to tuple-pinning is correct since this assertion will need updating ✓

### Strengths worth calling out

1. **The no-op prediction is preregistered as the falsifiable claim**, not a hidden assumption. This converts a "boring expected null" into a real test of the Sprint 37 engine-surface analysis.
2. **Terminology correction is handled cleanly** — the "non-ancestor" → "non-search-space" rename is explained once in the Sprint Theme note and applied consistently. The reasoning (`logged_position_distribution` *is* an ancestor of `policy_value` via the two-hop path; what matters for A1 binding is search-space membership) is correct.
3. **Implementation note about `bandit_log.py:540`** in the prompt (the list-comprehension pattern silently drops the new edge) is an exactly-right warning; without it, a literal-minded implementation could ship a broken graph that still passes the existing tests.
4. **Scope discipline** (one node, one edge, no engine change, no second slice) and the four-branch "What Happens After Sprint 38" both reduce the chance of mid-sprint scope creep — exactly the anti-pattern the Sprint 36 review caught.
5. **Path-by-path engine analysis** in §"What the engine would do with the widened graph" is the load-bearing argument for predicting H0, and every cited line checks out.

### Nits (skip if not iterating)

- The recommendation references "Sprint 38 tracked issue: `#201`" at the end; the prompt doesn't mention an issue number. Worth aligning if `#201` is canonical.
- Both docs are dense and repeat each other heavily. That's deliberate (the prompt is designed to be self-contained for an implementation agent), but the duplication makes the line-number drift in the prompt easier to miss in future edits. Consider extracting the engine-path table to a shared section that both docs link to, so it has one source of truth.

---

datablogin:
# Claude Code Review

## Review Feedback

Spot-checked the plan's claims against the actual code. All line numbers and function-behavior claims I verified resolve correctly:

- `bandit_log.py`: `propensity_schema` at 181 ✓, `self._position` at 195 ✓, `run_experiment` 346–369 with row-mask branch at 364–367 ✓, `get_prior_graph` 513–541 with list-comprehension at 540 ✓
- `suggest.py`: `_get_focus_variables` 1170–1180 ✓, `_apply_minimal_focus_a1` 1183–1236 with binding check at 1224 ✓, `_suggest_exploitation` starts at 532, parent-weighted block 604–619 ✓, `_rerank_alignment_only` 848–877 ✓, `_causal_alignment_score` 1015–1054 ✓
- `open_bandit_benchmark.py` line 541 `pomis_minimal_focus=(strategy == "causal")` ✓

The engine-path no-op argument is sound: under the widened graph, `_ancestors_in_space` still returns the 6 search-space vars (because it filters through `search_space.variable_names`), A1's binding condition `len(ancestors_in_space) == len(all_var_names)` still holds, `parents(policy_value)` is still 6 (the new edge targets `position_handling_flag`, not `policy_value`), and the alignment-score loop skips `logged_position_distribution` via `var_map.get(name) is None`.

## Overall

High-quality planning PR. The preregistered falsifiable no-op prediction is unusually disciplined — it converts "nothing changed" into a concrete falsification test of Sprint 37's engine-path analysis, which is the right stance. Internal consistency between the plan and the orchestration prompt (hard constraints, reproducibility command, H0/H1/H2, exit criteria) is tight. The "non-search-space" terminology retirement is a good catch and well-explained.

## Constructive feedback

**1. Path-4 filter description is a simplification of the real predicate.** The plan at "path 4" (in `_causal_alignment_score`) describes the filter as:

> `var = var_map.get(name); if var is None: continue` at lines 1044–1046

The actual predicate at line 1045 is:

```python
if var is None or var.lower is None or var.upper is None:
    continue
```

The simplified form reaches the right conclusion for `logged_position_distribution` (it hits the `var is None` branch because it's not in `var_map`), but a reader comparing line-by-line will notice the mismatch. Worth tightening to quote the actual predicate. A bonus observation this opens up: categorical variables (`position_handling_flag`) are *already* skipped in Sprint 37's alignment score via the `var.lower is None` branch — which strengthens, rather than weakens, the no-op claim and could be noted explicitly.

**2. Path-1 wording could be sharper.** The plan says `_get_focus_variables` "on its own remains a no-op — the full search-space fallback triggers." Reading 1176–1180, `focus` equals the full search space (6 vars), which is truthy and returned directly — the `focus if focus else search_space.variable_names` fallback does not actually trigger; the intersection is just equal to the full space. Functionally identical, but "fallback triggers" mis-describes the control flow. Suggest: "returns the full search space verbatim (the intersection equals the full space), so path 1 on its own is a no-op."

**3. Implementation note about the list comprehension is great — consider strengthening the test guidance.** The plan already says "pin edges by name tuple, not by counting `len(search_space.variables)`". Worth adding explicitly that the existing assertion `assert len(graph.edges) == 6` at `test_bandit_log_prior_graph.py:71` needs to become `== 7`, and that the ancestors test at `:92` needs to assert `set(_SEARCH_VARIABLES) | {"logged_position_distribution"}` rather than `set(_SEARCH_VARIABLES)`. That way the TDD step has a concrete before/after diff rather than a general direction.

**4. One wording nit in the widened-graph prediction.** The plan repeatedly says the widened graph is "predicted to be a no-op along every engine path that already existed under Sprint 37's minimal-focus wiring." Strictly speaking, the reason it is a no-op is that the new node is non-search-space *and* the new edge does not add a parent to `policy_value`. If either of those two conditions were relaxed (e.g., a future sprint adds an edge from the new node directly to `policy_value`), parent-weighting would change. Worth one sentence making that dual condition explicit, so a Sprint 39 reader doesn't transplant the no-op prediction to a different widening.

**5. Minor: exit-criterion and success-criterion overlap.** "Exit Criterion" items 1–5 and "Success Criteria" items 1–6 cover nearly identical ground. Not wrong, but could be consolidated or one could delete one and reference the other.

## Risk flags (none blocking)

- H0 is explicitly a null prediction at n=10 seeds. The plan's "Power and the H0-vs-H1 boundary" section is honest about the ~1% effect floor, and the Sprint 39 branch table names a diagnostic rerun for H2 before Sprint 39 decides direction. That is the right shape.
- No concerns about the Section 7 gate inheritance, permutation seed (`20260419`), or backend provenance (Ax/BoTorch only) — all consistent with Sprint 35/37.

Net: docs-only, preregistration-style plan. Accurate, internally consistent, appropriately narrow in scope. The tightening above would make it crisper but none of it is blocking for a planning PR.

---

datablogin:
# Claude Code Review

## Review Feedback

## Review: Sprint 38 Post-A1 Plan

This is a docs-only PR (planning + implementation prompt). My review focused on whether the line-level code references ground the engine-path no-op analysis accurately, whether the preregistered outcome is falsifiable, and whether the implementation contract the prompt prescribes is internally consistent.

### Verified code references

Every line-level reference I spot-checked is accurate. The plan pins its technical claims tightly to the current code:

| Plan claim | File:line | Verified |
|---|---|---|
| `run_experiment` 346–369, row-mask 364–367 | `bandit_log.py:346,364-367` | ✓ |
| `self._position` at line 195 | `bandit_log.py:195` | ✓ |
| `propensity_schema` ClassVar at 181 | `bandit_log.py:181` | ✓ |
| `get_prior_graph` 513–541, edge list at 540 | `bandit_log.py:513-541` | ✓ |
| `_get_focus_variables` 1170–1180 | `suggest.py:1170-1180` | ✓ |
| `_apply_minimal_focus_a1` 1183–1236 | `suggest.py:1183-1236` | ✓ |
| `_suggest_exploitation` starts 532, parent-weight 604–619 | `suggest.py:532,604-619` | ✓ |
| `use_soft` at 784, ancestor set 810–813, call at 839 | `suggest.py:784,810-813,839` | ✓ |
| `_rerank_alignment_only` 848–877 | `suggest.py:848-877` | ✓ |
| `_causal_alignment_score` 1015–1054, loop 1043–1052, guard at 1045 | `suggest.py:1015-1054` | ✓ |
| `pomis_minimal_focus=(strategy == "causal")` at line 541 | `open_bandit_benchmark.py:541` | ✓ |
| Test file lines 66, 71, 92, 94 | `test_bandit_log_prior_graph.py` | ✓ |

I also confirmed `CausalGraph.ancestors()` (`types.py:240`) performs a transitive BFS, so the plan's corrected claim that `logged_position_distribution` **is** an ancestor of `policy_value` via the two-hop path is graph-theoretically correct — that's what motivates the terminology switch from "non-ancestor" to "non-search-space", which is a genuine technical correction, not a cosmetic one.

### Engine-path no-op analysis is sound

Each of the four non-inert paths checks out under the widened graph:

1. **Path 1 `_get_focus_variables`:** filters ancestors through `search_space.variable_names`, so a non-search-space node is excluded by construction. Return value unchanged.
2. **Path 2 `_apply_minimal_focus_a1`:** binding condition is `len(ancestors_in_space) == len(all_var_names)`; since `logged_position_distribution` is excluded by `_ancestors_in_space`, both sides are unchanged, so A1 still binds and returns the same intersection.
3. **Path 3 parent-weighted exploitation:** uses `causal_graph.parents(objective_name)`, which is direct-parents only. The new edge targets `position_handling_flag`, so `parents(policy_value)` is unchanged, so `weights = None` still holds.
4. **Path 4 `_causal_alignment_score`:** the `var is None or var.lower is None or var.upper is None` guard at `suggest.py:1045` skips (a) the new non-search-space node (via `var_map.get(...) is None`) and (b) the categorical `position_handling_flag` (via `var.lower is None`). So `diffs` still has length 5, averaged over the same five continuous ancestors — the plan's observation here is correct and actually tightens the no-op claim.

### Strengths

- **Implementation trap call-out is sharp.** The prompt's "Implementation note" at section 1 correctly warns that `bandit_log.py:540` uses `edges = [(name, "policy_value") for name in self.get_search_space().variable_names]` — a comprehension pattern that **silently drops** the new non-search-space-sourced edge if an implementer naïvely extends the list. The prompt prescribes an explicit edge list as the fix. This is a real bug trap worth calling out.
- **Preregistered H0/H1/H2 is properly falsifiable.** The no-op is framed as the falsifiable claim, not as post-hoc justification. Assumptions (`causal_softness = 0.5`, `causal_exploration_weight = 0.0`, Ax/BoTorch available, `pomis_minimal_focus = True` on causal arm only, no engine-surface changes between Sprint 37 and 38) are enumerated.
- **Terminology note is a genuine correction,** not a rename — it aligns the engine-path reasoning with what `ancestors()` actually computes.
- **Exit criteria + "must not do" list are concrete,** issue-closable, and testable.

### Minor nits

1. **Prompt inconsistency at lines 99–104 vs 130–133** (`sprint-38-open-bandit-post-a1-follow-up.md`). The edge-count guidance says "rename the test to match" for `test_graph_has_six_directed_edges → test_graph_has_seven_directed_edges`, but the parallel guidance for the node count (`test_graph_has_seven_nodes → test_graph_has_eight_nodes`) omits the rename instruction. Both tests are named after counts, so both should be renamed for consistency. One-liner fix.

2. **Prompt item 2 is missing an explicit positive test for the new edge.** The prompt says to pin "the seven directed edges: the six Sprint 37 edges plus the new `("logged_position_distribution", "position_handling_flag")` edge", and to narrow/replace the Sprint 37 `test_no_chain_edges_between_search_variables` test. But there's no explicit prescription like `assert ("logged_position_distribution", "position_handling_flag") in graph.edges`. Relying on edge-count + a narrowed chain-edge test leaves a gap where a typo in the node name (e.g. `logged_position_distro`) could pass both checks. A dedicated single-edge membership assertion would close that gap.

3. **Plan line 364 wording** ("the ancestor set grows by one (`logged_position_distribution`), but that variable is not in the search space…"). Technically correct — Sprint 38's `ancestors(policy_value)` has one more element than Sprint 37's. The wording is fine; flagging only because "grows by one" appears alongside the broader "no-op" claim, and readers who skim might take those as contradictory. The paragraph resolves it, just worth a re-read in context.

4. **Nit on plan line 337 parenthetical** ("the `focus if focus else search_space.variable_names` guard at line 1180 does not trigger a fallback — it just returns the already-equal set"). This is correct but slightly subtle: `focus` is a *truthy non-empty list* under the widened graph, so the ternary returns `focus` (not the fallback). Fine as written, just worth noting that if a future graph widening accidentally filtered the search space down to empty, the fallback *would* silently hide the regression. Not a Sprint 38 concern, but a Sprint 39+ plan that widens with a more aggressive filter should re-verify.

### Verdict

The plan's empirical contract and engine-surface reasoning are tight, and the preregistered outcome is the right shape (no-op prediction as the falsifiable claim, not a justification). I'd merge once nits #1 and #2 above are addressed — both are in the prompt, not the plan proper, and both are one-liners.

---

datablogin:
# Claude Code Review

## Review Feedback

## Review: Sprint 38 Post-A1 Recommendation (Option B) — docs-only PR

Scope: two new files, 1,091 added lines. No code changes. I spot-checked the load-bearing code references and the test file the implementation prompt prescribes edits against. All references are accurate against the current tree.

### What works well

- **Preregistered no-op is a genuinely falsifiable claim, not a hedge.** The plan's central move — predicting that a structural widening is a no-op along every engine path that already existed, and declaring a B80 trending/certified row in either direction as falsification — is the right shape for a sprint after a null-result sprint. The per-path breakdown (paths 1–6) in "What the engine would do with the widened graph" makes the prediction verifiable by code inspection, not just by rerun.
- **Code-grounded edge justification.** The `logged_position_distribution -> position_handling_flag` edge is anchored to `bandit_log.py` lines 364–367 (verified: `if position_flag == "position_1_only": mask = self._position == 0`). That is the correct site — the row-mask branch that makes `self._position` (a data property, not a knob) structurally upstream of the flag.
- **Terminology correction handled cleanly.** The "non-ancestor" → "non-search-space" retitle in the Sprint Theme note is honest about the graph-theoretic imprecision introduced by the two-hop path. It's called out in both documents and doesn't silently rewrite prior sprint language.
- **Implementation prompt is precise and actionable.** The flagged pitfall at `bandit_log.py` line 540 (the list-comprehension `edges = [(name, "policy_value") for name in ...]` cannot express the new non-search-space source) is the exact bug that would silently drop the new edge. Calling it out in the prompt is load-bearing guidance.
- **Test-migration checklist is accurate.** I verified each cited line in `tests/unit/test_bandit_log_prior_graph.py`:
  - `test_graph_has_seven_nodes` count assert at line 66 ✓
  - `test_graph_has_six_directed_edges` count assert at line 71 ✓
  - `test_every_search_variable_is_an_ancestor_of_policy_value` at 86–92 ✓
  - `test_no_chain_edges_between_search_variables` at 94 with `assert v == "policy_value"` at line 100 — the prompt correctly notes this will break under the new edge.
- **Hard exclusions are concrete.** "No second new node" explicitly enumerates `request_item_overlap` as the one the Sprint 36 plan also listed, guarding against the exact scope-creep pattern Sprint 36 review caught.
- **Exit-criterion four-way branching** (exhausted-D / trending-diagnostic / certified-positive / certified-negative) each names a specific Sprint 39 move. No "we'll figure it out later."

### Minor issues

1. **Absolute local path in the reproducibility command.** Both files embed `/Users/robertwelborn/Projects/_local/causal-optimizer/...` in the `--data-path` and `--output` flags. Matches Sprint 37's convention so reproducibility-for-this-user is preserved, but another reviewer hitting the command out-of-the-box will need to edit two paths. Consider a one-line prelude noting "replace `_local/causal-optimizer` with your local OBD data root" — or accept that this is the established Sprint 35/37 pattern.

2. **Power caveat could be tighter.** The "Power and the H0-vs-H1 boundary" section notes that at n=10 per arm a ~1% causal advantage could land in H2 rather than H1. Fair — but H0 is also 10-seed-gated. If the true effect is null *and* the test is underpowered, H0 passes for the wrong reason. The plan partially addresses this by making Option D contingent on B20/B40 also being near-parity, but the H0-as-null-result asymmetry deserves one explicit sentence in the reconciliation section the Sprint 38 report must write.

3. **Test-file line references cite assertion lines, not `def` lines.** E.g., `test_graph_has_seven_nodes` is defined at line 61; the `assert len(graph.nodes) == 7` is at 66. The prompt cites 66. Accurate but mildly confusing — implementers scrolling to "line 66" will land inside the test body, not at its header. Not worth a rewrite, but an `ast-grep` or header-based citation would be friendlier.

4. **Path 4 reranker analysis is correct but dense.** The "the same guard already skips categorical variables — `var.lower is None` is true for `position_handling_flag`" aside is load-bearing (it strengthens rather than weakens the no-op claim) but buried in the middle of a long bullet. Lifting it to a short sentence of its own would make the argument easier to re-audit.

5. **Recommendation length vs. orchestration prompt overlap.** At 752 + 339 lines there is substantial overlap between the two files (H0/H1/H2, exclusion list, reproducibility command). The recommendation explicitly calls out that duplication ("this prompt duplicates those items so an implementation agent has a self-contained brief"), which is the right call — but when the implementation agent amends the prompt they must remember to amend the recommendation first. That convention is stated, but easy to miss. A single "if you edit anything here, edit 27-sprint-38-recommendation.md first and copy down" banner at the top of the prompt would be harder to miss than the current mid-document note.

### Consistency checks passed

- Engine surface claims (no edit to `suggest.py`, `loop.py`, benchmarks) match the Sprint 37 state verified by spot-checks at `suggest.py` lines 604–619, 784, 810–813, 839, 848–877, 1015–1054, 1170–1180, 1183–1236 and `open_bandit_benchmark.py` line 541.
- `propensity_schema` ClassVar claim at `bandit_log.py` line 181 verified (`propensity_schema: ClassVar[PropensitySchema] = PROPENSITY_SCHEMA_CONDITIONAL`).
- Null-control permutation seed `20260419` is consistent with Sprint 35/37.
- Two-sided MWU convention and scorecard bands (≤0.05 / 0.05–0.15 / >0.15) match Sprint 33/35/37.

### Bottom line

This is a well-constructed planning PR: preregistration is genuine, the falsifiability story is self-consistent, and the code references are accurate. The five review iterations show — this is in much tighter shape than most planning docs I see. The minor nits above are polish, not blockers.

Recommend merge after human sign-off, per the project's PR merge policy (claude-review.sh + greploop + human approval — per `CLAUDE.md`, do not merge autonomously).

---

